PR #23, opened 2018-08-10
greedy tree pruning

m09:
Greedy method for tree pruning that minimizes the number of comparisons to do at the cost of some more memory usage.

At this point there is some code duplication (_build_instances_index) because we are all working on the same functions and I didn't want to enter some merge hell. I'll PR a refactoring later on when more stable :)

### DIFF
--- a/lookout/style/format/rules.py
+++ b/lookout/style/format/rules.py
@@ -2,7 +2,7 @@ from collections import defaultdict
 from itertools import chain
 import functools
 import logging
-from typing import Union, List, Tuple, Dict, Iterable, NamedTuple
+from typing import Union, List, Tuple, Dict, Iterable, NamedTuple, Sequence, Set
 
 import modelforge
 import numpy
@@ -24,9 +24,8 @@ RuleAttribute = NamedTuple(
 RuleStats = NamedTuple("RuleStats", (("cls", int), ("conf", float)))
 """
 `cls` is the predicted class
-`conf` is the rule confidence \in [0, 1], "1" means super confident
+`conf` is the rule confidence \\in [0, 1], "1" means super confident
 """
-
 
 Rule = NamedTuple("RuleType", (("attrs", Tuple[RuleAttribute, ...]), ("stats", RuleStats)))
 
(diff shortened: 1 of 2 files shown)


m09:
Should be ready for another round of review @vmarkovtsev!

lookout/style/format/rules.py
```diff
@@ -57,11 +56,16 @@ class Rules(BaseEstimator, ClassifierMixin):
 
     CompiledRulesType = Dict[int, CompiledFeatureRules]
 
+    BottomUpGreedyBudget = NamedTuple("BottomUpGreedyBudget", (
```
vmarkovtsev: I see that the algorithm's name is "top-down", so makes sense to rename this to `TopDownGreedyBudget`

lookout/style/format/rules.py
```diff
@@ -252,11 +267,62 @@ def _compile_rules(cls, rules: List[Rule]) -> CompiledRulesType:
...
+    @classmethod
+    def _prune_branches_top_down_greedy(cls, rules: Sequence[Rule], X: numpy.ndarray,
+                                        Y: numpy.ndarray,
+                                        budget: BottomUpGreedyBudget) -> List[Rule]:
+        absolute, value = budget
+        if absolute:
+            assert isinstance(value, int)
+            n_budget = max(0, min(value, len(rules)))
+        else:
+            assert isinstance(value, float) and value >= 0 and value <= 1
```
vmarkovtsev: I think it can be `int` too in case it equals to 0 and 1

lookout/style/format/rules.py
```diff
@@ -57,11 +56,16 @@ class Rules(BaseEstimator, ClassifierMixin):
...
 
+    BottomUpGreedyBudget = NamedTuple("BottomUpGreedyBudget", (
+        ("absolute", bool), ("value", Union[float, int])))
+
     log = logging.getLogger("Rules")
 
     def __init__(self,
                  base_model: Union[DecisionTreeClassifier, RandomForestClassifier],
-                 prune_branches=True, prune_attributes=True, uncertain_attributes=True):
+                 prune_branches=True, prune_branches_algorithm="top-down-greedy",
```
zurk: I think we will have other pruning methods and it is better to create a namespace for its names, like:
```python
class PruneAlgorithms:
    top_down_greedy = "top-down-greedy"
```

vmarkovtsev: So far we have only two, let's YAGNI for now

zurk: I will add third one :)

zurk: but it is possible that for different scenarios, trees or data different methods can be more suitable. Or do I miss something?

vmarkovtsev: It is not clear now, but once we run experiments and see which pruning works the best (and in sane time) - we remove the rest.

m09: I didn't touch that @zurk since it was still debated. I also agree with @vmarkovtsev that once we'll have tested the 3 approaches a clear winner should emerge.

PR #27, opened 2018-08-13
Add the tool to generate language properties

vmarkovtsev:


vmarkovtsev:
@EgorBu So this should help with new languages. Unfortunately, I haven't found a reliable way to extract operators and reserved keywords too.

PR #32, opened 2018-08-14
Use sklearn's apply function to build instances index

m09:
I noticed that the apply function of sklearn was way faster than what I was using before to compute the rule to instances index I need. Here is a pull request to use it.

I also added a random forest to the tests.

### DIFF
--- a/lookout/style/format/rules.py
+++ b/lookout/style/format/rules.py
@@ -1,8 +1,8 @@
 from collections import defaultdict
-from itertools import chain, islice
+from itertools import islice
 import functools
 import logging
-from typing import Union, List, Tuple, Dict, Iterable, NamedTuple, Sequence, Set
+from typing import Union, List, Tuple, Dict, Iterable, Mapping, NamedTuple, Sequence, Set
 
 import modelforge
 import numpy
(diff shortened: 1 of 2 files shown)


lookout/style/format/rules.py
```diff
@@ -272,19 +281,26 @@ def _prune_branches(cls, rules: Sequence[Rule],
         # TODO(vmarkovtsev): implement this function
         return rules
 
-    @classmethod
-    def _build_instances_index(cls, rules: Sequence[Rule],
-                               X: numpy.ndarray) -> Dict[int, Set[int]]:
+    def _build_instances_index(self, rules: Sequence[Rule], X: numpy.ndarray,
+                               leaf2rule: Sequence[Mapping[int, int]]) -> Dict[int, Set[int]]:
+
+        self.log.debug('building instances index')
```
vmarkovtsev: Oops sorry missed the single quotes here

https://github.com/src-d/guide/blob/master/engineering/conventions-python.md#code-style

vmarkovtsev:
@m09 automerge failed here, can you please remove my merge attempt and rebase

PR #40, opened 2018-08-17
Improve sklearn compatibility

m09:
Use external functions instead of `get_params` and `_get_param_names` to decide which parameters to save in the `modelforge.Model`. Fit the base model if not fitted to allow for convenient CV workflows in sklearn.

### DIFF
--- a/lookout/style/format/rules.py
+++ b/lookout/style/format/rules.py
@@ -1,16 +1,17 @@
 from collections import defaultdict
-from itertools import islice
 import functools
+from itertools import islice
 import logging
-from typing import Union, List, Tuple, Dict, Iterable, Mapping, NamedTuple, Sequence, Set
+from typing import Any, Dict, Iterable, List, Mapping, NamedTuple, Sequence, Set, Tuple, Union
 
 import modelforge
 import numpy
-from sklearn.exceptions import NotFittedError
 from scipy.stats import fisher_exact
 from sklearn.base import BaseEstimator, ClassifierMixin
-from sklearn.tree import DecisionTreeClassifier, _tree as Tree
 from sklearn.ensemble import RandomForestClassifier
+from sklearn.exceptions import NotFittedError
+from sklearn.tree import _tree as Tree, DecisionTreeClassifier
+from sklearn.utils.validation import check_is_fitted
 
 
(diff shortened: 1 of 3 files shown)


lookout/style/format/rules.py
```diff
@@ -1,16 +1,17 @@
-from collections import defaultdict
-from itertools import islice
 import functools
 import logging
-from typing import Union, List, Tuple, Dict, Iterable, Mapping, NamedTuple, Sequence, Set
+from collections import defaultdict
+from itertools import islice
+from typing import Any, Dict, Iterable, List, Mapping, NamedTuple, Sequence, Set, Tuple, Union
```
vmarkovtsev: the order of imports should not depend on "from"

lookout/style/format/rules.py
```diff
@@ -201,13 +202,17 @@ def wrapped_check_fittable(self: "Rules", *args, **kwargs):
...
-        different from the ones the base model was trained on (actually, they should).
+        Trains the rules using the base tree model and the samples (X, y). If `base_model` is
+        already fitted, the samples may be different from the ones that were used.
 
         :param X: input features.
         :param y: input labels - the same length as X.
         :return: self
         """
+        try:
+            check_is_fitted(self.base_model, 'tree_')
```
vmarkovtsev: We should move this check to the constructor and always require the model to be fitted.

lookout/style/format/rules.py
```diff
@@ -242,6 +247,16 @@ def count_attrs():
         self._compiled = self._compile_rules(self._rules)
         return self
 
+    def get_saved_param_names(self) -> Sequence[str]:
+        param_names = self._get_param_names()
+        param_names.remove("base_model")
+        return param_names
+
+    def get_saved_params(self, deep: bool = False) -> Mapping[str, Any]:
```
vmarkovtsev: I suspect that `deep` is not needed here and we can safely hardcode `False`

lookout/style/format/rules.py
m09: I checked the fitting in the `fit` method to allow for idiomatic sklearn workflows, in particular CV. Requiring the base model to be fit in the constructor prevents from using almost all of sklearn.model_selection.

PR #41, opened 2018-08-17
Filter out buggy files during feature extraction

m09:
Some files are not successfully parsed with our current method. At this point we could either use them in an unstable state (features might not make sense) or discard them. This PR goes for the second option.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -1,4 +1,5 @@
 import importlib
+import logging
 from typing import Dict, Iterable, List, Mapping, NamedTuple, Optional, Sequence, Tuple
 
 import bblfsh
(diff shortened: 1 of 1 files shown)


lookout/style/format/features.py
```diff
@@ -99,6 +100,8 @@ class FeatureExtractor:
     feature_names = ["start_offset", "start_line", "start_col", "end_offset", "end_line",
                      "end_col", "internal_role"]
 
+    log = logging.getLogger("FeaturesExtractor")
```
vmarkovtsev: Pls rename it to `_log` - we should not allow externals to log under our name.

lookout/style/format/features.py
```diff
@@ -124,7 +127,12 @@ def extract_features(self, files: List[File]) -> Tuple[numpy.ndarray, numpy.ndar
         for file in files:
             contents = file.content.decode("utf-8", "replace")
             uast = file.uast
-            vnodes, parents = self._parse_file(contents, uast)
+            try:
+                vnodes, parents = self._parse_file(contents, uast)
+            except AssertionError as e:
+                self.log.warning("could not parse file %s reliably, skipping it", file.path,
```
vmarkovtsev: ~reliably~
skipping -> skipped

PR #48, opened 2018-08-21
Add reduced_error_prune method

zurk:
Move pruning method from the research part to the main code base.

### DIFF
--- a/lookout/style/format/rules.py
+++ b/lookout/style/format/rules.py
@@ -12,6 +12,7 @@ from scipy.stats import fisher_exact
 from sklearn.base import BaseEstimator, ClassifierMixin
 from sklearn.ensemble import RandomForestClassifier
 from sklearn.exceptions import NotFittedError
+from sklearn.metrics import accuracy_score
 from sklearn.tree import _tree as Tree, DecisionTreeClassifier
 from sklearn.utils.validation import check_is_fitted
 
@@ -273,6 +274,7 @@ class TrainableRules(BaseEstimator, ClassifierMixin):
                                  algorithm. Available for DecisionTreeClassifier model only.
         """
         super().__init__()
+        self.prune_branches_algorithm = prune_branches_algorithm
         self.base_model = base_model
         self.prune_branches = prune_branches
         self.top_down_greedy_budget = top_down_greedy_budget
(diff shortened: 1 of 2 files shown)


lookout/style/format/rules.py
```diff
@@ -158,7 +159,8 @@ def __init__(self,
                  base_model: Union[DecisionTreeClassifier, RandomForestClassifier],
                  prune_branches: bool = True, prune_branches_algorithm: str = "top-down-greedy",
                  top_down_greedy_budget: TopDownGreedyBudget = (False, 1.0),
-                 prune_attributes: bool = True, uncertain_attributes: bool = True):
+                 prune_attributes: bool = True, uncertain_attributes: bool = True,
+                 prune_base_model: bool = False):
```
vmarkovtsev: remove this please, we already have `prune_branches_algorithm`. We will never combine these two.

lookout/style/format/rules.py
```diff
@@ -390,11 +404,69 @@ def _compile_rules(cls, rules: Sequence[Rule]) -> CompiledRulesType:
...
+            parents.update(
+                {x: i for i, x in enumerate(tree.children_right) if x != Tree.TREE_LEAF})
+            return parents
+
     @classmethod
-    def _prune_branches(cls, rules: Sequence[Rule],
-                        X: numpy.ndarray, Y: numpy.ndarray) -> List[Rule]:
-        # TODO(vmarkovtsev): implement this function
-        return rules
+    def reduced_error_prune(cls, model: DecisionTreeClassifier, validation_X, validation_y,
```
vmarkovtsev: name: `_prune_reduced_error`

lookout/style/format/rules.py
```diff
@@ -390,11 +404,69 @@ def _compile_rules(cls, rules: Sequence[Rule]) -> CompiledRulesType:
...
+                    if verbose:
+                        print("Remove %d and %d leafs. Score change: %f" % (
+                            backup[0], backup[1], score_change))
+                    curent_score = new_score
+                    leafs.remove(second_leaf)
+                    leafs[leaf_index] = parent
+                    changes = True
+                    break
+        if verbose:
+            print("Finished.\n")
```
vmarkovtsev: We use logging everywhere

vmarkovtsev:
@m09 requesting your review here

lookout/style/format/rules.py
```diff
@@ -390,11 +404,69 @@ def _compile_rules(cls, rules: Sequence[Rule]) -> CompiledRulesType:
...
+            tree.children_right[node_to_prune] = Tree.TREE_LEAF
+            tree.feature[node_to_prune] = Tree.TREE_UNDEFINED
+            return backup
+
+        def _restore_tree(tree, node, backup):
+            tree.children_left[node], tree.children_right[node], tree.feature[node] = backup
+
+        model = deepcopy(model)
+        tree = model.tree_
+        init_score = curent_score = model.score(validation_X, validation_y)
```
m09: `curent_score` should be `current_score`

lookout/style/format/rules.py
```diff
@@ -390,11 +404,69 @@ def _compile_rules(cls, rules: Sequence[Rule]) -> CompiledRulesType:
...
+        changes = True
+        bad_change = set()
+        parents = cls._get_parents(tree)
+        leafs = list(numpy.where(tree.children_left == Tree.TREE_LEAF)[0])
```
m09: `leafs` should be `leaves`

lookout/style/format/rules.py
```diff
@@ -390,11 +404,69 @@ def _compile_rules(cls, rules: Sequence[Rule]) -> CompiledRulesType:
...
+                second_leaf = tree.children_right[parent]
+                second_leaf = second_leaf if second_leaf != leaf else tree.children_left[parent]
+                if tree.children_left[second_leaf] != Tree.TREE_LEAF or \
+                        tree.children_right[second_leaf] != Tree.TREE_LEAF:
+                    continue
+
+                backup = _prune_tree(tree, parent)
+                new_score = model.score(validation_X, validation_y)
+                score_change = (new_score - curent_score) / curent_score
+                full_score_change = (new_score - init_score) / init_score
```
m09: It should be possible to compute the score change from `tree.n_node_samples` and `tree.value` which would avoid another pass over the data (and keeping this way to perform a test of the validity of the faster method would be great!).

lookout/style/format/rules.py
zurk: I need spell checker from @irinakhismatullina as soon as possible :sweat_smile:

lookout/style/format/rules.py
zurk: yeah, it is true. I think to optimize this part, and I decided not to do it now. Just I case we evaluate our pruning methods and exclude this one, no reason for that. So I leave it as is for now. 

But if you think it is better to do it now, I can

m09: yup I think it might be necessary just because we can't easily test it on what we'll use it on without this optimization and it will be interesting to see if this method performs better than the greedy one on "real" data (big repos)

zurk:
please wait for method optimization before next review

lookout/style/format/rules.py
vmarkovtsev: cc @eiso

vmarkovtsev:
@zurk I added a conflict, can you please resolve

zurk:
@vmarkovtsev @m09 ready for next review round.

lookout/style/format/rules.py
```diff
@@ -266,13 +268,15 @@ def __init__(self,
...
+                                     certainty (see "Generating Production Rules From Decision \
                                      Trees" by J.R. Quinlan).
+        :param prune_base_model: indicates whether to prune base_model via reduced error pruning \
+                                 algorithm. Available for DecisionTreeClassifier model only.
         """
         super().__init__()
+        self.prune_branches_algorithm = prune_branches_algorithm
         self.base_model = base_model
         self.prune_branches = prune_branches
-        self.prune_branches_algorithm = prune_branches_algorithm
```
vmarkovtsev: This swap of lines is not necessary: `self.prune_branches_algorithm = prune_branches_algorithm`

lookout/style/format/rules.py
```diff
@@ -291,6 +295,11 @@ def base_model(self, value: Union[DecisionTreeClassifier, RandomForestClassifier
             check_is_fitted(value, "tree_")
         elif isinstance(value, RandomForestClassifier):
             check_is_fitted(value, "estimators_")
+        if self.prune_branches_algorithm == "reduced-error" and \
+                isinstance(value, RandomForestClassifier):
+            raise ValueError("A reduced-error pruning method can be used only for "
```
vmarkovtsev: This is fine: you iterate over estimators and prune them independently.

lookout/style/format/rules.py
```diff
@@ -419,11 +430,75 @@ def _merge_rules(cls, rules: List[Rule]) -> List[Rule]:
             new_rules.append(Rule(tuple(new_rule), stats))
         return new_rules
 
+    @staticmethod
+    def _get_parents(tree: Tree.Tree) -> Dict[int, int]:
```
vmarkovtsev: You use `_get_parents` only once and the function is small, better to inline it.

lookout/style/format/rules.py
zurk: but as I remember from your experiment, it makes things worse. 

Should I include it anyway?

zurk:
Oh, sorry, @m09 I forget to make a comment. `n_node_samples` and `value` available only for the training dataset and as we know we should use another data part here. I can not recalculate it easily for new data, however `model.decision_path(X)` is available. So I take this way.

lookout/style/format/rules.py
vmarkovtsev: Yeah, please include this even though it works bad.

m09:
Good point, I didn't think of that. I guess a slightly faster way would be to use `apply`, compute stats for leaves (number of instances of each class in leaves) and then propagate to each non-terminal node, which would be very fast. It would allow not to recompute the sum of instances which are of a given class when we prune repeatedly the same part of the tree. But I think it's even more premature to optimize now in this way than in my last comment so we should wait IMO :)

zurk:
mmm, yeah, did not think about `apply `. Ok.

PR #67, opened 2018-08-27
Provide a config file option to tweak loggers output levels.

m09:
Make fine-grained logging output tweaks easy by providing a way to set levels through conf files.

### DIFF
--- a/lookout/core/cmdline.py
+++ b/lookout/core/cmdline.py
@@ -64,7 +64,7 @@ def run_analyzers(args):
     :param args: Parsed command line arguments.
     :return: None
     """
-    slogging.setup(args.log_level, args.log_structured)
+    slogging.setup(args.log_level, args.log_structured, args.log_config_path)
     log = logging.getLogger("run")
     model_repository = create_model_repo_from_args(args)
     log.info("Created %s", model_repository)
@@ -96,7 +96,7 @@ def init_repo(args):
     :param args: Parsed command line arguments.
     :return: None
     """
-    slogging.setup(args.log_level, False)
+    slogging.setup(args.log_level, False, args.log_config_path)
     repo = create_model_repo_from_args(args)
     repo.init()
 
@@ -121,9 +121,11 @@ def add_model_repository_args(parser):
                help="Additional keyword arguments to SQLAlchemy database engine.")
 
 
-def add_log_level_arg(parser):
+def add_logging_args(parser):
     parser.add("--log-level", default="INFO", choices=logging._nameToLevel,
                help="Logging verbosity.")
+    parser.add("--log-config-path",
+               help="Path to the file which sets individual log levels of domains.")
 
 
 def create_parser():
@@ -143,7 +145,7 @@ def create_parser():
     run_parser = add_parser(
         "run", "Launch a new service with the specified (one or more) analyzers.")
     run_parser.set_defaults(handler=run_analyzers)
-    add_log_level_arg(run_parser)
+    add_logging_args(run_parser)
     run_parser.add("--log-structured", action="store_true",
                    help="Enable structured logging (compatible with k8s).")
     run_parser.add("-c", "--config", is_config_file=True,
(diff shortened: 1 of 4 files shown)


lookout/core/cmdline.py
```diff
@@ -64,7 +64,7 @@ def run_analyzers(args):
     :param args: Parsed command line arguments.
     :return: None
     """
-    slogging.setup(args.log_level, args.log_structured)
+    slogging.setup(args.log_config_path, args.log_level, args.log_structured)
```
vmarkovtsev: I want to make the configuration file optional - so better to put it to the last place with an empty default value.

lookout/core/slogging.py
```diff
@@ -110,10 +116,16 @@ def ensure_utf8_stream(stream):
 
     sys.stdout, sys.stderr = (ensure_utf8_stream(s)
                               for s in (sys.stdout, sys.stderr))
-    logging.basicConfig(level=level)
+
+    logging.basicConfig()
+    if os.path.exists(config_path):
+        with open(config_path, 'r') as fh:
```
vmarkovtsev: you don't need `'r'` here - it is by default

lookout/core/tests/test_slogging.py
```diff
@@ -11,7 +11,7 @@ class SloggingTests(unittest.TestCase):
     def test_structured_logging(self):
         logging.basicConfig()
         handler_backup = logging.getLogger().handlers[0]
-        slogging.setup("INFO", True)
+        slogging.setup("logging.yaml", "INFO", True)
```
vmarkovtsev: I think the order is different here so Travis failed

lookout/core/slogging.py
```diff
@@ -89,14 +92,17 @@ def flush(self):
         sys.stdout.flush()
 
 
-def setup(level: Union[str, int], structured: bool):
+def setup(level: Union[str, int], structured: bool, config_path: str):
```
vmarkovtsev: Please give it an empty value by default

lookout/core/slogging.py
```diff
@@ -110,10 +114,16 @@ def ensure_utf8_stream(stream):
...
-    logging.basicConfig(level=level)
+
+    logging.basicConfig()
+    if config_path:
+        with open(config_path) as fh:
+            config = yaml.safe_load(fh)
+        for key, val in config.items():
+            logging.getLogger(key).setLevel(logging._nameToLevel.get(val, level))
     root = logging.getLogger()
-    # In some cases root has handlers and basicConfig() is a no-op
```
vmarkovtsev: This comment should be preserved, otherwise people will decide to optimize and remove "root.setLevel(level)" - I personally did that once to my own code and left that comment after.

lookout/core/cmdline.py
```diff
@@ -121,9 +121,10 @@ def add_model_repository_args(parser):
                help="Additional keyword arguments to SQLAlchemy database engine.")
 
 
-def add_log_level_arg(parser):
+def add_logging_args(parser):
     parser.add("--log-level", default="INFO", choices=logging._nameToLevel,
                help="Logging verbosity.")
+    parser.add("--log-config-path", help="Path of the logging config.")
```
vmarkovtsev: Path **to** the file which sets individual log levels of domains.

lookout/core/slogging.py
m09: I removed the comment because before it was basicConfig(level=level) but now I call it without level and set level afterwards. It is now way less tempting to optimize but I can put it back anyway :)

vmarkovtsev: Good catch! We need to retain `basicConfig(level=level)` then because the root is not always set in the configuration file.

m09: we set it afterwards so it's ok. Isn't it?

vmarkovtsev: Yes indeed. We definitely need to comment on that :)

PR #74, opened 2018-08-28
Add no-op nodes to be able to predict from and to lack of whitespace

### LINKED ISSUE #72: Create NOOP class and related nodes
At this point, we do not have a way to predict whitespace from a lack of whitespace. We also don't have a way to predict "no whitespace" from whitespace. The NOOP class will serve this purpose and will have associated nodes in-between every current node so that we are able to predict those instances.

m09:
Fixes #72.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -94,7 +94,6 @@ class VirtualNode:
                                        col=node.end_position.col))
 
 
-CLS_NONE = "<none>"
 CLS_SPACE = "<space>"
 CLS_SPACE_INC = "<+space>"
 CLS_SPACE_DEC = "<-space>"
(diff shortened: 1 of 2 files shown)


lookout/style/format/features.py
```diff
@@ -104,8 +104,9 @@ def from_node(node: bblfsh.Node, file: str) -> Iterable["VirtualNode"]:
 CLS_NEWLINE = "<newline>"
 CLS_SINGLE_QUOTE = "'"
 CLS_DOUBLE_QUOTE = '"'
+CLS_NOOP = "<noop>"
```
vmarkovtsev: We need to remove `CLS_NONE` - it is a duplicate

lookout/style/format/features.py
```diff
@@ -289,6 +291,16 @@ def _classify_vnodes(self, nodes: Iterable[VirtualNode]) -> List[VirtualNode]:
                             y=cls))
         return result
 
+    @staticmethod
+    def _add_noops(vnodes: Sequence[VirtualNode]) -> List[VirtualNode]:
```
vmarkovtsev: docstring missing

vmarkovtsev:
I really wonder how our quality will drop after this change :smile:

m09:
I'm running some tests to find out but they are at least slower if not worse in quality ><

m09:
So on freecodecamp quality went from 95.3% to 99.4%. It seems noop is extremely easy to predict and even helps with other classes for some reason (otherwise accuracy would have jumped only to ~97.6). I'll try on more repos to try and see why this happens.

vmarkovtsev:
We have unexpectedly reached @EgorBu's accuracy. We need to change the target metric in this case.

PR #78, opened 2018-08-29
Extend roles to nodes non-handled by bblfsh

m09:
We discussed a while ago with @vmarkovtsev that we should extend roles to include tokens that babelfish doesn't handle out of the box.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -322,7 +322,7 @@ class FeatureExtractor:
                               y=CLASS_INDEX[CLS_NOOP])]
         for vnode in vnodes:
             result.append(vnode)
-            result.append(VirtualNode(value="", start=vnode.start, end=vnode.end,
+            result.append(VirtualNode(value="", start=vnode.end, end=vnode.end,
                                       y=CLASS_INDEX[CLS_NOOP]))
         return result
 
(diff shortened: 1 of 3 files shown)


lookout/style/format/tests/test_features.py
```diff
@@ -116,6 +117,14 @@ def test_noop_vnodes(self):
             else:
                 self.assertEqual(vnode.y, CLASS_INDEX[CLS_NOOP])
 
+    def test_extended_roles(self):
+        file = File(content=bytes(self.contents, 'utf-8'),
+                    uast=self.uast)
+        X, _, _ = self.extractor.extract_features([file])
+        # last columns are only roles
+        last_columns = self.extractor.parents_depth + self.extractor.siblings_window
+        self.assertGreater(numpy.count_nonzero(X[:, -last_columns:] > len(ROLE_INDEX)), 0)
```
vmarkovtsev: We also return vnodes corresponding to X. Can you please add the check for vnode.value in RESERVED_INDEX

m09: Actually we only have roles for the neighbours of the vnode of `X` (the role of `X` itself being the target), so it might be hard and `vnode.value` might not be in `RESERVED_INDEX` for legitimate reasons.

m09:
I'm on it but I observe some inconsistent features so I'm debugging. I'll PR as soon as I have found the cause of the inconsistencies.

m09:
@vmarkovtsev pong

lookout/style/format/features.py
```diff
@@ -322,7 +322,7 @@ def _add_noops(vnodes: Sequence[VirtualNode]) -> List[VirtualNode]:
                               y=CLASS_INDEX[CLS_NOOP])]
         for vnode in vnodes:
             result.append(vnode)
-            result.append(VirtualNode(value="", start=vnode.start, end=vnode.end,
+            result.append(VirtualNode(value="", start=vnode.end, end=vnode.end,
```
vmarkovtsev: Good one

PR #82, opened 2018-08-30
[WIP] Configure documentation

m09:
Sphinx configuration to allow for ReadTheDocs webhook to build documentation on push.

PR #83, opened 2018-08-31
Configure documentation

### LINKED ISSUE #71: Build and publish the documentation to ReadTheDocs
We need to build the documentation from docstrings and publish it somewhere.

I guess there are no alternatives to Sphinx (though I hate it) and ReadTheDocs.

cc/@smola

m09:
Sphinx configuration to allow for ReadTheDocs webhook to build documentation on push. Fixes #71. Continuation of #82.


vmarkovtsev:
@m09 as far as I remember Sphinx imports modules. Can you also add a check in the CI that the docs are buildable - say, in the same job as we check flake8

m09:
I use some new stuff that doesn't import modules (https://sphinx-autoapi.readthedocs.io/en/latest/) which is way better because the build was killed on RTD for eating too much RAM when it was trying to pip install . to be able to import modules, so this alternative is welcome.

PR #91, opened 2018-09-02
Update lookout binary from 0.1.0 to 0.1.3 to match submodule version and update related tests

m09:
The two latest `lookout` submodule updates didn't bump the binary used to perform the tests. Here is the bump.

### DIFF
--- a/lookout/core/tests/server.py
+++ b/lookout/core/tests/server.py
@@ -17,7 +17,7 @@ def fetch():
     try:
         buffer = io.BytesIO()
         with urlopen("https://github.com/src-d/lookout/releases/download/"
-                     "v0.1.0/lookout_sdk_v0.1.0_linux_amd64.tar.gz") as response:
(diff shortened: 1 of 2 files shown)


lookout/core/tests/server.py
```diff
@@ -17,7 +17,7 @@ def fetch():
     try:
         buffer = io.BytesIO()
         with urlopen("https://github.com/src-d/lookout/releases/download/"
-                     "v0.1.0/lookout_sdk_v0.1.0_linux_amd64.tar.gz") as response:
+                     "v0.1.3/lookout_sdk_v0.1.3_linux_amd64.tar.gz") as response:
```
vmarkovtsev: We need to add the submodule hash here in the future. It is possible to fetch a tarball for any specific revision.

m09: :+1:. #95.

PR #104, opened 2018-09-06
Skip noop features

m09:
The skipping part is implemented in this PR but it uncovered some other bug that I'll investigate later. I'll ping when it's fixed. Update: other bugs fixed in #110.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -435,24 +435,42 @@ class FeatureExtractor:
                         break
                     current_ancestor = parents[current_ancestor_id]
 
-            col_offset = 0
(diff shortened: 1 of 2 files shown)


m09:
yes it is.

lookout/style/format/features.py
```diff
@@ -435,24 +435,36 @@ def _inplace_write_vnode_features(
                         break
                     current_ancestor = parents[current_ancestor_id]
 
-            col_offset = 0
+            start_left = i - 1
```
vmarkovtsev: Would be nice to have a comment which explains this dark `+-=1` magic :)

m09: done

Issue #103, opened 2018-09-06
`analyzer list` is broken


EgorBu:
Hi,
functionality to provide list of analyzers is broken.
```console
egor@egor-sourced:~/workspace/style-analyzer$ analyzer list
/usr/local/lib/python3.5/dist-packages/sklearn/ensemble/weight_boosting.py:29: DeprecationWarning: numpy.core.umath_tests is an internal NumPy module and should not be imported. It will be removed in a future NumPy release.
  from numpy.core.umath_tests import inner1d
Traceback (most recent call last):
  File "/usr/local/bin/analyzer", line 11, in <module>
    load_entry_point('lookout-style', 'console_scripts', 'analyzer')()
  File "/home/egor/workspace/style-analyzer/lookout/__main__.py", line 16, in main
    return handler(args)
  File "/home/egor/workspace/style-analyzer/lookout/core/cmdline.py", line 44, in list_analyzers
    m = importer.find_module(name).load_module(name)
  File "<frozen importlib._bootstrap_external>", line 388, in _check_name_wrapper
  File "<frozen importlib._bootstrap_external>", line 809, in load_module
  File "<frozen importlib._bootstrap_external>", line 668, in load_module
  File "<frozen importlib._bootstrap>", line 268, in _load_module_shim
  File "<frozen importlib._bootstrap>", line 693, in _load
  File "<frozen importlib._bootstrap>", line 673, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 665, in exec_module
  File "<frozen importlib._bootstrap>", line 222, in _call_with_frames_removed
  File "/home/egor/workspace/style-analyzer/lookout/style/format/__init__.py", line 1, in <module>
    from lookout.style.format.analyzer import FormatAnalyzer
  File "/home/egor/workspace/style-analyzer/lookout/style/format/__init__.py", line 1, in <module>
    from lookout.style.format.analyzer import FormatAnalyzer
  File "/home/egor/workspace/style-analyzer/lookout/style/format/analyzer.py", line 16, in <module>
    from lookout.style.format.model import FormatModel
  File "/home/egor/workspace/style-analyzer/lookout/style/format/model.py", line 8, in <module>
    from lookout.style.format.rules import Rules, Rule, RuleAttribute, RuleStats
  File "/home/egor/workspace/style-analyzer/lookout/style/format/rules.py", line 168, in <module>
    class TrainableRules(BaseEstimator, ClassifierMixin):
[...]

PR #105, opened 2018-09-06
Fix for different behaviour of `typing` in python3.5 & python3.6+

EgorBu:
Fix for https://github.com/src-d/style-analyzer/issues/103

### DIFF
--- a/lookout/core/analyzer.py
+++ b/lookout/core/analyzer.py
@@ -1,4 +1,4 @@
-from typing import NamedTuple, Any, Dict  # noqa: F401
+from typing import NamedTuple, Any, Dict
 
 from modelforge import Model
 


lookout/style/format/rules.py
```diff
@@ -281,8 +281,7 @@ def base_model_name(self) -> str:
         return self._base_model_name
 
     @base_model_name.setter
-    def base_model_name(self, value: Union[str, Type[DecisionTreeClassifier],
-                                           Type[RandomForestClassifier]]):
+    def base_model_name(self, value: Union[str, DecisionTreeClassifier, RandomForestClassifier]):
```
vmarkovtsev: This contradicts with `issubclass` check on line 293. We need to go deeper.

lookout/style/format/rules.py
```diff
@@ -5,6 +5,36 @@
 import logging
 from typing import (Any, Dict, Iterable, List, Mapping, NamedTuple, Optional, Sequence, Set, Tuple,
                     Union, Type)
+import sys
+
+if sys.version_info[:2] == (3, 5):
```
vmarkovtsev: We import `Type` in many places. This should into `lookout/core/ports.py`

PR #50, opened 2018-08-21
Refactored typos analyzer added

irinakhismatullina:
Signed-off-by: Irina Khismatullina <irenekhismatullina@gmail.com>

PR #106, opened 2018-09-06
Add utils for typos correction model

irinakhismatullina:
First part of typos analyzer 
1. Added dependencies to requirements.txt
2. Added common functions used for typos correction on different stages
3. Added classes and utils for candidates generation and candidates ranking

### DIFF
--- a/lookout/style/format/tests/test_analyzer.py
+++ b/lookout/style/format/tests/test_analyzer.py
@@ -58,6 +58,7 @@ class AnalyzerTests(unittest.TestCase):
             cls.uast = bblfsh.Node.FromString(fin.read())
         cls.base_files = cls.get_files_from_tar(str(base / "freecodecamp-base.tar.xz"))
         cls.head_files = cls.get_files_from_tar(str(base / "freecodecamp-head.tar.xz"))
+        cls.ptr = ReferencePointer("someurl", "someref", "somecommit")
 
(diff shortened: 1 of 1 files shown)


zurk:
Is it the first part of https://github.com/src-d/style-analyzer/pull/50? 

And a tip about commits messages: we do not use past simple and first word should be a verb, so something like: 
```
Add typos correcting model dependencies to requirements.txt
Add typos correction utils
Add typos correction candidates generation
Add candidates ranking
```

would be great! 

Also, one more PRO tip: if work in progress and PR is not ready for a review you can always add [WIP] to PR name. I am not sure if it is a case this time.

irinakhismatullina:
Yes, it is first part, so it is ready for review.

OK, thanks, will do that in future. Should I edit commit messages now too?

zurk:
I think it is better to edit. it does not take a lot of time ;)

m09:
Hey Irina, thanks for refactoring, it looks like you've already improved things a lot :+1: 

It would be better to split this PR as follows:

1. symspell submodule
2. utils
3. ranking
4. candidates generator
5. typos corrector

Please name each PR something like `Add ... from #106 (x/5)`

I ordered the parts to follow the dependencies in your code this way you can always run the code you add, add tests, etc.

This will make the merging much faster.

PR #112, opened 2018-09-07
Add path to VirtualNode 

EgorBu:
It should help to debug and visualize predictions.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -30,7 +30,7 @@ class FeatureType(enum.Enum):
 
 class VirtualNode:
     def __init__(self,  value: str, start: Position, end: Position,
-                 *, node: bblfsh.Node = None, y: int = None):
+                 *, node: bblfsh.Node = None, y: int = None, path: str = None):
         """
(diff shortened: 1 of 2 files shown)


lookout/style/format/features.py
```diff
@@ -30,14 +30,15 @@ class FeatureType(enum.Enum):
 
 class VirtualNode:
     def __init__(self,  value: str, start: Position, end: Position,
-                 *, node: bblfsh.Node = None, y: int = None):
+                 *, node: bblfsh.Node = None, y: int = None, path: str = None):
```
m09: Should be compulsory argument if we set it everywhere and rely on it later on.

EgorBu: TBH - it used to work before without it. I'm using it for analysis - I'm not sure if it should be mandatory

vmarkovtsev: @EgorBu Do you mean we do not access this attribute anywhere in the code?

EgorBu: Done.

Issue #92, opened 2018-09-03
[style-analyzer] Find right balance to one hot encode roles


m09:
So at this point roles are handled as ordinal data which is not allowing for good splits on roles. We should find a way to one-hot encode them. Just straight-up one-hot encoding is maybe not optimal (130+ classes) but can be an option. It should be compared with smaller number of buckets (~10, ~50) if the performance is not great.

m09:
I thought a bit about it and even with sparse matrices, we'll have to take care not to end up with too many features for the tree. At this point there is around ~130 possible roles and this gets multiplied by all the siblings (left, parents, right) ie, with default parameters, by 12.

This means that if we don't bucket we have around 1500 features just for the roles. In my experience this is way too many.

zurk:
it is a lot I agree. I think once we definitely should make embeddings for Roles.

m09:
Yes if we are ok with having an external training phase it would be great to replace those with small embeddings (dimension 10 max?) to keep the feature space reasonnable.

vmarkovtsev:
Once we have embeddings we will not be able to explicitly express the learned rules, no?

Otherwise it is very easy to train - they are graph embeddings of each node. Not for Q3 though - too little time.

m09:
It's true that it's not interpretable.

zurk:
I think that we should have a pretraining model for each language we analyze, so there is no additional time cost for one more training. 
So, it is something like take a PGA -- > get role co-occurrence matrix --> run swivel.

To get co-occurrence matrix we can do the same thing we do for identifiers: add +1 co-occurrence for Roles in one scope. It is easy to code, but a lot of time to get the model. So I think for now it is more like enhancement direction.

zurk:
actually it is interpretable. It each leaf we will have the subset of Roles. The same thing we have now I beleve.

m09:
it is not interpretable in the sense that you can't modify the rules since they are operating in embedded space.

vmarkovtsev:
We can design the optimal bucketing scheme by clustering the roles embedding space.

zurk:
I make experiments with one hot encoding preprocessing and here is the first results:
1. BayesSearchCV gives a weird error if I use sparse matrix. So, It just cannot handle it. To test everything I convert it to regular ndarray (yeap, memory inefficient). I set `n_iter=10` 
2. I use `freecodecamp` data from tests to test the performance.  
[...]

PR #116, opened 2018-09-10
Add roles one hot encoding option

zurk:
Work in progress.
Related to https://github.com/src-d/style-analyzer/issues/92


lookout/style/format/features.py
```diff
@@ -248,9 +435,28 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None
...
-    def _classify_vnodes(self, nodes: Iterable[VirtualNode], path: str) -> Iterable[VirtualNode]:
+    def encode_roles(self, X: numpy.ndarray) -> coo_matrix:
+        """
+        Encodes all roles features in data matrix X to its one hot encoding representation.
+        Fits the model if it is unfitted.
+        :param X: Data matrix.
+        :return: New sparse data matrix.
+        """
+        # It is not possible to have negative values to run OneHotEncoder, so here is dirty hack
+        X += 1
```
m09: This dirty hack works since we use trees and they are indifferent to strictly monotonic transformations of their features but it would still be good to actually only +1 the categorical columns so that we can interpret stuff later on. It would even be better to not touch anything and just modify tokens.py and roles.py to start at 1 instead of 0.

zurk: > since we use trees and they are indifferent to strictly monotonic transformations

Yup, that why I use it. I like the last suggestion! Will implement

zurk:
as soon as I should do the same for new features, I close this PR, wait for https://github.com/src-d/style-analyzer/pull/135 to be merged. Make more experiments for bucketing and PR the best approach.

PR #117, opened 2018-09-18
Add length feature to left and right siblings

m09:
Length feature doesn't leak information about the current node but can still help with the formatting problem.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -171,9 +171,9 @@ class FeatureExtractor:
              FeaturesGroup(("start_line", "start_col"), None)),
             (FeatureType.left_siblings,
              FeaturesGroup(("start_line_diff", "end_line_diff", "start_col_diff", "end_col_diff",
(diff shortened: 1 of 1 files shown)


lookout/style/format/features.py
```diff
@@ -401,10 +401,13 @@ def _get_left_sibling_features(self, left_sibling_vnode: VirtualNode, vnode: Vir
                 abs(left_sibling_vnode.end.line - vnode.end.line),
                 left_sibling_vnode.start.col - vnode.start.col,
                 left_sibling_vnode.end.col - vnode.end.col,
+                left_sibling_vnode.end.offset - left_sibling_vnode.start.offset,
```
EgorBu: isn't it a duplication of ```left_sibling_vnode.start.col - vnode.start.col, left_sibling_vnode.end.col - vnode.end.col```?

m09: No, this is the relative position of the neighbor compare to the node while my addition is just the length of the neighbor.

PR #119, opened 2018-09-18
Add debugging capabilities to features.py

m09:
Small PR to allow for debugging of features.py. It's what I've been using to browse the different errors we get during parsing.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -43,6 +43,7 @@ class VirtualNode:
         self.value = value
         assert start.line >= 1 and start.col >= 1, "start line and column are 1-based like UASTs"
         assert end.line >= 1 and end.col >= 1, "end line and column are 1-based like UASTs"
+        assert y in EMPTY_CLS or start.offset < end.offset, "illegal empty node"
         self.start = start
         self.end = end
         self.node = node
@@ -146,12 +147,15 @@ CLS_NOOP = "<noop>"
 CLASSES = (CLS_SPACE, CLS_TAB, CLS_NEWLINE, CLS_SPACE_INC, CLS_SPACE_DEC,
            CLS_TAB_INC, CLS_TAB_DEC, CLS_SINGLE_QUOTE, CLS_DOUBLE_QUOTE, CLS_NOOP)
 CLASS_INDEX = {cls: i for i, cls in enumerate(CLASSES)}
+EMPTY_CLS = frozenset([CLASS_INDEX[CLS_TAB_DEC], CLASS_INDEX[CLS_SPACE_DEC],
+                       CLASS_INDEX[CLS_NOOP]])
 
 
 class FeatureExtractor:
     _log = logging.getLogger("FeaturesExtractor")
 
-    def __init__(self, language: str, *, siblings_window: int = 5, parents_depth: int = 2):
+    def __init__(self, language: str, *, siblings_window: int = 5, parents_depth: int = 2,
+                 debug_parsing: bool = False):
         """
         Construct a `FeatureExtractor`.
 
@@ -160,6 +164,7 @@ class FeatureExtractor:
         """
         self.siblings_window = siblings_window
         self.parents_depth = parents_depth
+        self.debug_parsing = debug_parsing
         language = language.lower()
         self.tokens = importlib.import_module("lookout.style.format.langs.%s.tokens" % language)
         self.roles = importlib.import_module("lookout.style.format.langs.%s.roles" % language)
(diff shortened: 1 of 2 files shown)


m09:
This should be merged after the next one that I push btw, not now.

lookout/style/format/features.py
```diff
@@ -43,6 +43,7 @@ def __init__(self,  value: str, start: Position, end: Position,
         self.value = value
         assert start.line >= 1 and start.col >= 1, "start line and column are 1-based like UASTs"
         assert end.line >= 1 and end.col >= 1, "end line and column are 1-based like UASTs"
+        assert y in EMPTY_CLS or start.offset < end.offset, "illegal empty node"
```
EgorBu: Could you clarify this line? Without context, I can't understand why it's required

lookout/style/format/features.py
```diff
@@ -83,6 +84,12 @@ def from_node(node: bblfsh.Node, file: str, path: str) -> Iterable["VirtualNode"
                               node=node, path=path)
             return
         start_offset = outer_token.find(node.token)
+        assert start_offset >= 0, (
+            "Couldn't find the token in the specified position:\nNode role: %s\nParsed form: “%s”"
```
EgorBu: File information is needed

lookout/style/format/features.py
```diff
@@ -232,9 +243,12 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None
             try:
                 vnodes, parents = self._parse_file(contents, uast, file.path)
             except AssertionError as e:
-                self._log.warning("could not parse file %s with error \"%s\', skipping",
+                self._log.warning("could not parse file %s with error '%s', skipping",
                                   file.path, e)
-                continue
+                if self.debug_parsing:
```
EgorBu: Alternative is to use https://github.com/src-d/ml/blob/69361199e04f393d1a10b10badb969a3f5b2ed24/sourced/ml/utils/engine.py#L83-L92

lookout/style/format/features.py
```diff
@@ -146,12 +153,15 @@ def to_comment(self, correct_y: int) -> Comment:
 CLASSES = (CLS_SPACE, CLS_TAB, CLS_NEWLINE, CLS_SPACE_INC, CLS_SPACE_DEC,
            CLS_TAB_INC, CLS_TAB_DEC, CLS_SINGLE_QUOTE, CLS_DOUBLE_QUOTE, CLS_NOOP)
 CLASS_INDEX = {cls: i for i, cls in enumerate(CLASSES)}
+EMPTY_CLS = frozenset([CLASS_INDEX[CLS_TAB_DEC], CLASS_INDEX[CLS_SPACE_DEC],
```
EgorBu: why do you need to extract hash value from `EMPTY_CLS`? Or you want to make this structure unmodifiable?

lookout/style/format/features.py
```diff
@@ -232,9 +243,12 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None
             try:
                 vnodes, parents = self._parse_file(contents, uast, file.path)
             except AssertionError as e:
-                self._log.warning("could not parse file %s with error \"%s\', skipping",
+                self._log.warning("could not parse file %s with error '%s', skipping",
                                   file.path, e)
-                continue
+                if self.debug_parsing:
+                    input("press Enter to continue")
```
vmarkovtsev: How about `import ipdb; ipdb.set_trace()` then?

lookout/style/format/features.py
m09: It's a test that empty nodes are only of the classes from which we expect empty nodes (-space, -tab and noop)

lookout/style/format/features.py
m09: file information is already present where all the asserts are caught

lookout/style/format/features.py
m09: true but I feel like it's more verbose, needs an import and doesn't filter on exception type so maybe it's better to stick to that

lookout/style/format/features.py
m09: it's a reflex: when I use something primarily for `in` operations I use sets. And this won't be modified so `frozenset` in this case.

m09:
This one should pass travis after #123 merge

lookout/style/format/features.py
m09: I went for printing the trace directly instead, it proved to be the fatest during debugging. I can change if needed.

Issue #120, opened 2018-09-18
Fix config inconsistencies


m09:
While trying to apply external config to an analyzer, I noticed several inconsistent points:

- in [manager.py](https://github.com/src-d/style-analyzer/blob/master/lookout/core/manager.py#L55), we use the class name of the analyzer to request a config object to the lookout server but the analyzer is known by its module to the lookout server. So we'll ask for the config of `FormatAnalyzer` but it'll be under the `lookout.style.format` key for the lookout server.

- in [analyzer.py](https://github.com/src-d/style-analyzer/blob/master/lookout/style/format/analyzer.py), we retrieve config in two different ways in `train` and `analyze`: in `analyze`, the config is in the attributes of the objects itself while in `train` it is passed explicitely to the function. That makes it hard to reuse initialization functions such as [_load_train_config](https://github.com/src-d/style-analyzer/blob/master/lookout/style/format/analyzer.py#L142) and sure enough we don't reuse it, which means we can end up with different values for each config parameter during the training and the eval, which is super bad. (btw, even without this problem we might end up with different values since they are not model-bound at this point.)

vmarkovtsev:
According to [PEP 3155](https://www.python.org/dev/peps/pep-3155/) we can use `__qualname__` to fix the first point, BUT you need to check how it works with plain file names - we support them besides imports.

As for the second point, `analyze()` is an instance method as supposedly review events are more frequent than pushes. So you can override the class constructor.

m09:
`__qualname__` is not usable I think since with the `analyzer_class` mechanism we can end up with classes in different depths. Ie here it's in `lookout.style.format.analyzer` but in some other package it might have been in `lookout.style.format.core.analyzer`, which means we can't just take qualname and remove a level.

PR #121, opened 2018-09-18
Streamline config in analyzers

m09:
Uses a dict in both analyze and train.

### DIFF
--- a/lookout/core/manager.py
+++ b/lookout/core/manager.py
@@ -53,8 +53,10 @@ class AnalyzerManager(EventHandlers):
         for analyzer in self._analyzers:
             try:
                 mycfg = dict(request.configuration[analyzer.__name__])
+                self._log.info("%s config: %s", analyzer.__name__, mycfg)
             except (KeyError, ValueError):
                 mycfg = {}
+                self._log.debug("no config was provided for %s", analyzer.__name__)
             model, cache_miss = self._model_repository.get(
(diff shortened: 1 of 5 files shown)


m09:
Related to #120 but doesn't fix it all, only the second point.

lookout/core/manager.py
```diff
@@ -53,8 +53,10 @@ def process_review_event(self, request: ReviewEvent) -> EventResponse:
         for analyzer in self._analyzers:
             try:
                 mycfg = dict(request.configuration[analyzer.__name__])
+                self._log.info("config: %s", mycfg)
             except (KeyError, ValueError):
                 mycfg = {}
+                self._log.info("no config provided")
```
vmarkovtsev: Please change to debug log level

vmarkovtsev: Also message:

`"no config was provided for %s", analyzer.__name__`

lookout/core/manager.py
```diff
@@ -53,8 +53,10 @@ def process_review_event(self, request: ReviewEvent) -> EventResponse:
         for analyzer in self._analyzers:
             try:
                 mycfg = dict(request.configuration[analyzer.__name__])
+                self._log.info("config: %s", mycfg)
```
vmarkovtsev: `"%s config: %s", analyzer.__name__, mycfg`

PR #122, opened 2018-09-18
Fix config finding

m09:
Edit of the core package to also keep track of the module name to be able to request the correct config.

### DIFF
--- a/lookout/core/doc/getting_started.md
+++ b/lookout/core/doc/getting_started.md
@@ -65,9 +65,10 @@ class MyModel(AnalyzerModel):
 class MyAnalyzer(Analyzer):
     model_type = MyModel
     version = "1"
+    name = "my_analyzer.MyAnalyzer"
     description = "Reports the changes in UAST node counts."
     _log = logging.getLogger("MyAnalyzer")
-    
+
     @with_changed_uasts_and_contents
     def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
                 data_request_stub: DataStub, changes: Iterable[Change]) -> [Comment]:
@@ -128,7 +129,7 @@ lookout push ipv4://localhost:2000 --git-dir /tmp/go-git \
     --to 4397264e391b45a8eac147cc7373189d55c640cc
 ```
 
-You should have `/tmp/tmp/go-git/MyAnalyzer_1.asdf`.
+You should have `/tmp/tmp/go-git/my_analyzer.MyAnalyzer_1.asdf`.
 
 Run the model:
 
--- a/lookout/core/analyzer.py
+++ b/lookout/core/analyzer.py
@@ -40,7 +40,7 @@ class AnalyzerModel(Model):
         :return: self
         """
         assert isinstance(self, analyzer.model_type)
-        self.name = analyzer.__name__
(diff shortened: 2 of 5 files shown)


lookout/core/cmdline.py
```diff
@@ -75,7 +75,7 @@ def run_analyzers(args):
     data_service = DataService(data_request_address)
     log.info("Created %s", data_service)
     manager = AnalyzerManager(
-        analyzers=[importlib.import_module(a).analyzer_class for a in args.analyzer],
+        analyzers={a: importlib.import_module(a).analyzer_class for a in args.analyzer},
```
vmarkovtsev: Module names seem too hacky here. Besides, we can import the same modules by different paths and screw up the configuration. The better solution is to require `NAME` for every `Analyzer` - like we do in Modelforge models. That `NAME` should uniquely identify the class, can include all the qualification, etc.

m09: I pushed an edit implementing `name`. Should I also edit the cli so that it receives this name and not the module name we use currently? Pros: consistent, cons: longer to type

vmarkovtsev: The cli logic is simple and straightforward: it imports the module and takes the class. Nobody cares about the names except the developers who want to get the right config in their analyzer.

vmarkovtsev:
Missing: 

- update of https://github.com/src-d/style-analyzer/blob/master/lookout/core/doc/getting_started.md
- conflicts

lookout/core/doc/getting_started.md
```diff
@@ -117,7 +118,7 @@ objects with UASTs and file contents. If you need only the UASTs, use
 Launch the analyzer:
 
 ```
-analyzer run my_analyzer -c config.yml
+analyzer run lookout.mypackage.my_analyzer.MyAnalyzer -c config.yml
```
vmarkovtsev: This should work as in the old version.

lookout/core/doc/getting_started.md
```diff
@@ -65,9 +65,10 @@ class MyModel(AnalyzerModel):
 class MyAnalyzer(Analyzer):
     model_type = MyModel
     version = "1"
+    name = "lookout.mypackage.my_analyzer.MyAnalyzer"
```
vmarkovtsev: We should drop `lookout` - it is our official branded package root.

lookout/core/doc/getting_started.md
m09: good catch I edited too quickly

m09:
Not related but I have to relaunch plenty of jobs recently due to:

```
Collecting pyspark==2.2.1 (from sourced-engine<0.7,>=0.5.1->sourced-ml<0.7,>=0.6.0->lookout-style==0.0.2)
Could not install packages due to an EnvironmentError: [Errno 13] Permission denied: '/home/travis/.cache/pip/wheels/18/b3/89/7576d7fdc6434d76de21b06d6fbcc0bf136d681be2ff64e6af/pyspark-2.2.1-py2.py3-none-any.whl'
Consider using the `--user` option or check the permissions.
```

Like 1 or 2 jobs per PR, it's super annoying. Any idea why travis fails only some % of the time?

PR #123, opened 2018-09-19
Fix empty node creation

m09:
Fix the 2 last spots I found of buggy node creation. One of them I fixed with ad-hoc hack since I found no general solution that handles all cases.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -2,7 +2,8 @@ from collections import OrderedDict
 import enum
 import importlib
 import logging
-from typing import Dict, Iterable, List, Mapping, NamedTuple, Optional, Sequence, Tuple, Set
+from typing import (Callable, Dict, Iterable, List, Mapping, NamedTuple, Optional, Sequence, Set,
+                    Tuple)
 
 import bblfsh
 import numpy
@@ -66,13 +67,16 @@ class VirtualNode:
                self.path == other.path
 
     @staticmethod
-    def from_node(node: bblfsh.Node, file: str, path: str) -> Iterable["VirtualNode"]:
+    def from_node(node: bblfsh.Node, file: str, path: str,
+                  token_unwrappers: Mapping[str, Callable[[str], Tuple[str, str]]]
+                  ) -> Iterable["VirtualNode"]:
         """
         Initializes the VirtualNode from a UAST node. Takes into account prefixes and suffixes.
 
         :param node: UAST node
         :param file: the file contents
         :param path: the file path
+        :param token_unwrappers: mapping from bblfsh internal types to functions to unwrap tokens
         :return: new VirtualNode-s
         """
         outer_token = file[node.start_position.offset:node.end_position.offset]
@@ -82,7 +86,16 @@ class VirtualNode:
                               Position(*[f[1] for f in node.end_position.ListFields()]),
                               node=node, path=path)
             return
-        start_offset = outer_token.find(node.token)
+        node_token = node.token
+        if node.internal_type in token_unwrappers:
+            node_token, outer_token = token_unwrappers[node.internal_type](outer_token)
+        start_offset = outer_token.find(node_token)
+        assert start_offset >= 0, (
+            "Couldn't find the token in the specified position:\nNode role: %s\nParsed form: “%s”"
+            "\nRaw form: “%s”\nStart position: %d, %d, %d\nEnd position: %d, %d, %d" % (
+                node.internal_type, node_token, outer_token, node.start_position.offset,
+                node.start_position.line, node.start_position.col, node.end_position.offset,
+                node.end_position.line, node.end_position.col))
         start_pos = node.start_position.offset + start_offset
         if start_offset:
             yield VirtualNode(outer_token[:start_offset],
@@ -93,9 +106,9 @@ class VirtualNode:
                                        line=node.start_position.line,
                                        col=node.start_position.col + start_offset),
                               path=path)
-        end_offset = start_offset + len(node.token)
-        end_pos = start_pos + len(node.token)
-        yield VirtualNode(node.token,
+        end_offset = start_offset + len(node_token)
+        end_pos = start_pos + len(node_token)
+        yield VirtualNode(node_token,
                           Position(offset=start_pos,
                                    line=node.start_position.line,
                                    col=node.start_position.col + start_offset),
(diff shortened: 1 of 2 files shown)


lookout/style/format/features.py
```diff
@@ -82,7 +85,16 @@ def from_node(node: bblfsh.Node, file: str, path: str) -> Iterable["VirtualNode"
                               Position(*[f[1] for f in node.end_position.ListFields()]),
                               node=node, path=path)
             return
-        start_offset = outer_token.find(node.token)
+        token = node.token
```
vmarkovtsev: @m09 Can you please describe the problem which is solved here and give a few examples.

For sure this complicates the code so I need to figure out if there is a better fix.

m09: Sure. I'll start by describing the way we handle delimiters atm. When babelfish returns a `token` that is smaller that the piece of code spanned by the position of its node, we use the parts of the piece of code not covered by the node token as delimiters. That includes:

- `//` for `// It's a comment` (` It's a comment` is the token)
- `/*` and `*/` for `\* this is a block comment */` (` this is a block comment ` is the token)
- `"` and `"` for `"I'm a string"` (`I'm a string` is the token)

Now, sometimes, babelfish interprets what's in the delimiters. For example:

- `"a\"b"` will return `a"b` as a token, which means we can't find it directly in the piece of code covered by the node (the `\` is in the way)
- `1e2` will return `100.0` as a token (or `100` I'm not sure) which means once again we can't use the token information.

So it seems that we can't both have a general purpose delimiter finder based on the babelfish token and something that handles the case that interests us the most (strings, quotes are amongst our features) easily.

vmarkovtsev: Daaaaaaamn

vmarkovtsev: @juanjux @dennwc Is this smth you can easily fix or it comes directly from the native driver?

m09: just a clarification, I said the values were sometimes evaluated but they're actually always evaluated for SOME types, I'm not sure I was clear. So `StringLiteral` and `NumericLiteral` are always evaluated. Also, I found some instances of regexes that were parsed as strings (I'll try to reproduce if there's interest).

juanjux:
We definitely don't make these kind of translations so they must come from the native driver.

lookout/style/format/features.py
```diff
@@ -66,7 +67,9 @@ def __eq__(self, other: "VirtualNode") -> bool:
                self.path == other.path
 
     @staticmethod
-    def from_node(node: bblfsh.Node, file: str, path: str) -> Iterable["VirtualNode"]:
+    def from_node(node: bblfsh.Node, file: str, path: str,
+                  interpreted_types: Mapping[str, Callable[[str, str], Tuple[str, str]]]
+                  ) -> Iterable["VirtualNode"]:
```
vmarkovtsev: Docstring for `interpreted_types` missing

lookout/style/format/features.py
vmarkovtsev: Please rename to `node_token`

lookout/style/format/features.py
```diff
@@ -82,7 +85,16 @@ def from_node(node: bblfsh.Node, file: str, path: str) -> Iterable["VirtualNode"
                               Position(*[f[1] for f in node.end_position.ListFields()]),
                               node=node, path=path)
             return
-        start_offset = outer_token.find(node.token)
+        token = node.token
+        if node.internal_type in interpreted_types:
+            token, outer_token = interpreted_types[node.internal_type](token, outer_token)
```
vmarkovtsev: Looks like we never need `token` in the function call - it is interpreted so we never trust it - and this argument can be removed

lookout/style/format/features.py
```diff
@@ -163,6 +175,9 @@ def __init__(self, language: str, siblings_window: int = 5, parents_depth: int =
         language = language.lower()
         self.tokens = importlib.import_module("lookout.style.format.langs.%s.tokens" % language)
         self.roles = importlib.import_module("lookout.style.format.langs.%s.roles" % language)
+        self.interpreted_types = importlib.import_module(
+            "lookout.style.format.langs.%s.interpreted_types" % language) \
+            .INTERPRETED_TYPES
```
vmarkovtsev: We can remove the new line here, the whole thing is 91 chars long

lookout/style/format/langs/javascript/interpreted_types.py
```diff
@@ -0,0 +1,16 @@
+from typing import Tuple
+
+
+def preprocess_string_literal(token: str, outer_token: str) -> Tuple[str, str]:
```
vmarkovtsev: preprocess -> unwrap

lookout/style/format/langs/javascript/interpreted_types.py
```diff
@@ -0,0 +1,16 @@
...
+    assert len(outer_token) >= 2, "string literal should have at least 2 characters"
+    return outer_token[1:-1], outer_token
+
+
+def preprocess_use_outer_token(token: str, outer_token: str) -> Tuple[str, str]:
+    return outer_token, outer_token
+
+
+INTERPRETED_TYPES = {
```
vmarkovtsev: TOKEN_UNWRAPPERS

lookout/style/format/features.py
vmarkovtsev: + we need to catch ImportError here - there are some sane language drivers which do not do the shit

lookout/style/format/features.py
```diff
@@ -163,6 +176,12 @@ def __init__(self, language: str, siblings_window: int = 5, parents_depth: int =
         language = language.lower()
         self.tokens = importlib.import_module("lookout.style.format.langs.%s.tokens" % language)
         self.roles = importlib.import_module("lookout.style.format.langs.%s.roles" % language)
+        try:
+            self.token_unwrappers = importlib.import_module(
+                "lookout.style.format.langs.%s.token_unwrappers" % language).TOKEN_UNWRAPPERS
+        except ImportError:
+            # It's normal for some languages not to have a token_unwrappers module.
+            pass
```
vmarkovtsev: We should init `self.token_unwrappers` to empty dict here otherwise we hit AttributeError

lookout/style/format/features.py
```diff
@@ -163,6 +176,13 @@ def __init__(self, language: str, siblings_window: int = 5, parents_depth: int =
         language = language.lower()
         self.tokens = importlib.import_module("lookout.style.format.langs.%s.tokens" % language)
         self.roles = importlib.import_module("lookout.style.format.langs.%s.roles" % language)
+        try:
+            self.token_unwrappers = importlib.import_module(
+                "lookout.style.format.langs.%s.token_unwrappers" % language).TOKEN_UNWRAPPERS
+        except ImportError:
+            # It's normal for some languages not to have a token_unwrappers module.
+            self.token_unwrappers = {}
+            pass
```
vmarkovtsev: `pass` is not needed anymore

PR #126, opened 2018-09-20
Exclude garbage from lookout data requests

m09:
Uses the lookout server exclude pattern and mirrors this behavior for debugging tools in a separate function. Also filters on max line length.

### DIFF
--- a/lookout/core/garbage_exclusion.py
+++ b/lookout/core/garbage_exclusion.py
@@ -1,0 +1,15 @@
+"""Pattern to filter out garbage files."""
+from importlib import import_module
+from pathlib import Path
+from typing import Iterable
+
+
+def _gather_patterns() -> Iterable[str]:
+    """Return available patterns to filter out garbage files from the langs package."""
+    for path in (Path(__file__).parent / "langs").iterdir():
+        if path.is_dir():
+            if (path / "garbage.py").is_file():
+                yield import_module("lookout.core.langs.%s.garbage" % path.name).GARBAGE_PATTERN
+
+
+GARBAGE_PATTERN = "|".join(_gather_patterns())
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -42,8 +42,10 @@ class FormatAnalyzer(Analyzer):
         """
         comments = []
         changes = list(data["changes"])
-        base_files = self._files_by_language(c.base for c in changes)
-        head_files = self._files_by_language(c.head for c in changes)
+        base_files = self._files_by_language(self._filter_files((c.base for c in changes),
+                                                                self.config["line_length_limit"]))
+        head_files = self._files_by_language(self._filter_files((c.head for c in changes),
+                                                                self.config["line_length_limit"]))
         for lang, lang_head_files in head_files.items():
             try:
                 rules = self.model[lang]
@@ -90,7 +92,8 @@ class FormatAnalyzer(Analyzer):
         config = cls._load_config(config)
         cls.log.info("train %s %s %s", ptr.url, ptr.commit,
                      pformat(config, width=4096, compact=True))
-        files_by_language = cls._files_by_language(data["files"])
+        files_by_language = cls._files_by_language(cls._filter_files(data["files"],
+                                                                     config["line_length_limit"]))
         model = FormatModel().construct(cls, ptr)
         for language, files in files_by_language.items():
             language = language.lower()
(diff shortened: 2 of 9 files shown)


lookout/core/garbage_exclusion.py
```diff
@@ -0,0 +1 @@
+GARBAGE_PATTERN = r"^.*[.-]min\.js"
```
vmarkovtsev: We should gather this pattern by "|"-ing the patterns for each supported language.

lookout/style/format/analyzer.py
```diff
@@ -166,10 +169,26 @@ def _load_config(cls, config: Mapping[str, Any]):
...
+    def _filter_files(cls, files: Iterable[File], line_length_limit: int) -> Iterable[File]:
+        """
+        Filter files based on their maximum line length.
+
+        :param files: Files to filter.
+        :param line_length_limit: Maximum line length to accept a file.
+        :return: Files filtered.
+        """
+        for file in files:
+            text = file.content.decode("utf-8", "replace")
```
vmarkovtsev: We have https://docs.python.org/3.5/library/stdtypes.html#bytes.splitlines in 3.5+ so no need to decode

lookout/style/format/files_filtering.py
```diff
@@ -0,0 +1,28 @@
+"""Facilities to filter files during debugging."""
+from os.path import isfile
+import re
+from typing import Iterable, Optional
+
+
+def filter_filepaths(filepaths: Iterable[str], line_length_limit: int = 500,
+                     exclude_pattern: Optional[str] = r"[.-]min\.js$") -> Iterable[str]:
```
vmarkovtsev: We should remove the default value of `exclude_pattern`

lookout/style/format/files_filtering.py
```diff
@@ -0,0 +1,28 @@
...
+    :return Iterable of paths, filtered.
+    """
+    exclude_compiled_pattern = re.compile(exclude_pattern) if exclude_pattern else None
+    for filepath in filepaths:
+        if not isfile(filepath):
+            continue
+        if exclude_compiled_pattern and exclude_compiled_pattern.search(filepath):
+            continue
+        try:
+            with open(filepath, 'r') as fh:
```
vmarkovtsev: `max(fh, key=len)`

lookout/style/format/analyzer.py
vmarkovtsev: `max(text.splitlines(), key=len)`

vmarkovtsev: So we sacrifice `rstrip()` and accurate utf-8 decoding which is not really needed here

lookout/style/format/files_filtering.py
vmarkovtsev: `open(filepath, 'rb')`

lookout/core/garbage_exclusion.py
m09: I'll join a list of patterns but separating them explicitely in different languages shouldn't be done IMO to reflect the fact that lookout cannot filter based on language.

vmarkovtsev: Sure. This is done to explicitly force new language developers to think about exclusion patterns and define them in each language package

lookout/core/garbage_exclusion.py
```diff
@@ -0,0 +1,5 @@
+GARBAGE_PATTERNS = [
+    r"^.*[.-]min\.js"
```
vmarkovtsev: So this should go to `langs.javascript` package and loaded from all the languages

m09: As I mentioned in your last comment lookout doesn't differentiate by language so I think it's better to have a single list to be mindful of the interactions of the patterns for all languages at the same time.

lookout/style/format/files_filtering.py
```diff
@@ -0,0 +1,33 @@
...
+    for filepath in filepaths:
+        if not isfile(filepath):
+            continue
+        if exclude_compiled_pattern and exclude_compiled_pattern.search(filepath):
+            continue
+        try:
+            with open(filepath, 'rb') as fh:
+                if max(fh, key=len, default=0) <= line_length_limit:
+                    yield filepath
+        except UnicodeDecodeError:
```
vmarkovtsev: We will never have UnicodeDecodeError here thanks to `'rb'`

lookout/core/garbage_exclusion.py
vmarkovtsev: I don't want to update the global file for every language, this breaks the abstraction.

We do not care that lookout does not differentiate by language here at all.

1. get the list of packages under `lookout.style.format.langs`
2. take `garbage` module from each, take `GARBAGE_PATTERNS` from that module
3. `GARBAGE_PATTERNS = list(chain.from_iterable(...))`

m09: I'm obviously fine to do that but that means we'll have to handle those garbage files for every analyzer while we said that we should remove them globally. If we changed our mind I'll implement this analyzer contained version.

vmarkovtsev: The global `GARBAGE_PATTERNS` is still here in the same place in the global scope. This scanning code goes upper, that's all.

m09: Ok. I misunderstood because of the 1. that mentioned `lookout.style.format`. I'll put that in `lookout.core.langs` then.

lookout/core/garbage_exclusion.py
```diff
@@ -0,0 +1,15 @@
+"""Pattern to filter out garbage files."""
+from importlib import import_module
+from pathlib import Path
+from typing import Iterable
+
+
+def _gather_patterns() -> Iterable[str]:
+    """Return available patterns to filter out garbage files from the langs package."""
+    for path in (Path(__file__).parent / "langs").iterdir():
+        if path.is_dir():
```
vmarkovtsev: Since we are using `iterdir()`, do we need to further check `is_dir()`?

m09: yes it returns all files in a dir

PR #127, opened 2018-09-20
Transform debugging scripts so that they can be used as libs easily

m09:


lookout/style/format/__main__.py
```diff
@@ -48,12 +48,13 @@ def main():
...
     try:
         handler = args.handler
+        delattr(args, "handler")
     except AttributeError:
         def print_usage(_):
             parser.print_usage()
 
         handler = print_usage
-    return handler(args)
+    return handler(**vars(args))
```
vmarkovtsev: I don't see where `vars` comes from...

m09: it's to transform the `argsparse.Namespace` into a dict

zurk: cool!

PR #135, opened 2018-09-21
Dense one-hot encoding of the internal types

EgorBu:
Add OHE Roles & split internal types and keywords into 2 cols

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -11,6 +11,7 @@ import numpy
 from lookout.core.api.service_analyzer_pb2 import Comment
 from lookout.core.api.service_data_pb2 import File
 
+
 Position = NamedTuple("Position", (("offset", int), ("line", int), ("col", int)))
 """
 `line` and `col` are 1-based to match UAST!
@@ -195,9 +196,11 @@ class FeatureExtractor:
              FeaturesGroup(("start_line", "start_col"), None)),
             (FeatureType.left_siblings,
              FeaturesGroup(("start_line_diff", "end_line_diff", "start_col_diff", "end_col_diff",
-                            "length", "role_id"), siblings_window)),
+                            "length", "internal_type_id", "keyword_id", *self.roles.ROLES),
+                           siblings_window)),
             (FeatureType.right_siblings,
-             FeaturesGroup(("length", "role_id"), siblings_window)),
+             FeaturesGroup(("length", "internal_type_id", "keyword_id", *self.roles.ROLES),
+                           siblings_window)),
             (FeatureType.parents,
              FeaturesGroup(("role_id", ), parents_depth)),
         ])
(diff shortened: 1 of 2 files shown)


lookout/style/format/features.py
```diff
@@ -11,6 +11,9 @@
 from lookout.core.api.service_analyzer_pb2 import Comment
 from lookout.core.api.service_data_pb2 import File
 
+N_ROLES = len(bblfsh.aliases.DESCRIPTOR.enum_types_by_name["Role"].values_by_number.items())
+ROLES = [bblfsh.role_name(role) for role in range(N_ROLES)]
```
vmarkovtsev: Since we merged @zurk's PR, we should use the data which he collected

lookout/style/format/features.py
```diff
@@ -410,16 +414,28 @@ def _add_noops(vnodes: Sequence[VirtualNode], path: str) -> List[VirtualNode]:
...
 
+    def _get_keyword_index(self, vnode: VirtualNode) -> int:
+        keyword_index = -1
+        if not vnode.node and vnode.value in self.tokens.RESERVED_INDEX:
+            keyword_index = self.tokens.RESERVED_INDEX[vnode.value]
+        return keyword_index
+
+    @staticmethod
+    def _get_role_indices(vnode: VirtualNode) -> int:
+        role_indices = [0] * N_ROLES
```
vmarkovtsev: pls change to `numpy.zeros`

lookout/style/format/features.py
```diff
@@ -410,16 +414,28 @@ def _add_noops(vnodes: Sequence[VirtualNode], path: str) -> List[VirtualNode]:
...
         return role_index
 
+    def _get_keyword_index(self, vnode: VirtualNode) -> int:
+        keyword_index = -1
+        if not vnode.node and vnode.value in self.tokens.RESERVED_INDEX:
+            keyword_index = self.tokens.RESERVED_INDEX[vnode.value]
+        return keyword_index
+
+    @staticmethod
+    def _get_role_indices(vnode: VirtualNode) -> int:
```
vmarkovtsev: Update the return tye - it is different from int

lookout/style/format/features.py
EgorBu: current logic will not work with numpy arrays: `*self._get_role_indices(left_sibling_vnode)` this is how features are given to `_get_left_sibling_features` & `_get_right_sibling_features`

lookout/style/format/features.py
zurk: It is right. Only 88 roles are used in JavaScript
https://doc.bblf.sh/languages.html

vmarkovtsev: So instead of using a global variable, take it for each language

vmarkovtsev: See https://github.com/src-d/style-analyzer/blob/master/lookout/style/format/features.py#L182

PR #147, opened 2018-09-24
Refactor debugging facilities into a separate dir and add training script

m09:
The format package had almost more debugging modules than normal ones so I moved everything to a debug package. I also added a training function since we all have it duplicated in our research code otherwise.

### DIFF
--- a/lookout/style/format/debug/train.py
+++ b/lookout/style/format/debug/train.py
@@ -1,0 +1,45 @@
+"""Facilities to train a FormatModel without a lookout server."""
+from os.path import join
+from typing import Iterable
+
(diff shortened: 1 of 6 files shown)


lookout/style/format/debug/train.py
```diff
@@ -0,0 +1,45 @@
+"""Facilities to train a FormatModel without a lookout server."""
```
vmarkovtsev: Please put this file into `research` for now. I need to find some time to refactor it and make nice - but not this week.

m09:
Ok, let's just add some research module. Follows in #149.

PR #148, opened 2018-09-24
Add missing parameters for TrainableRules fit

zurk:
It is a bug. We did not pass all config parameters for final training that we use for `BayesSearchCV`.

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -153,8 +153,11 @@ class FormatAnalyzer(Analyzer):
             trainable_rules = TrainableRules(
                 prune_branches_algorithms=config["prune_branches_algorithms"],
                 prune_attributes=config["prune_attributes"],
-                random_state=config["random_state"],
-                uncertain_attributes=config["uncertain_attributes"], **bscv.best_params_
+                top_down_greedy_budget=config["top_down_greedy_budget"],
+                uncertain_attributes=config["uncertain_attributes"],
+                prune_dataset_ratio=config["prune_dataset_ratio"],
+                n_estimators=config["n_estimators"],
+                random_state=config["random_state"], **bscv.best_params_
             )
             trainable_rules.fit(X, y)
             model[language] = trainable_rules.rules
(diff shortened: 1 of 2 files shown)


zurk:
Now I see that we do not use `n_estimators` at all. 
Will fix it.

lookout/style/format/analyzer.py
```diff
@@ -153,8 +153,11 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_request_st
             trainable_rules = TrainableRules(
                 prune_branches_algorithms=config["prune_branches_algorithms"],
                 prune_attributes=config["prune_attributes"],
-                random_state=config["random_state"],
-                uncertain_attributes=config["uncertain_attributes"], **bscv.best_params_
+                top_down_greedy_budget=config["top_down_greedy_budget"],
```
vmarkovtsev: Can you please reorder this change so that there is only 1 removed line here (leave `random_state` and `uncertain` intact)

lookout/style/format/rules.py
```diff
@@ -217,11 +217,10 @@ def fit(self, X: numpy.ndarray, y: numpy.ndarray) -> "TrainableRules":
         :param y: input labels - the same length as X.
         :return: self
         """
-        base_model = self._base_model_class(max_depth=self.max_depth,
-                                            max_features=self.max_features,
-                                            min_samples_leaf=self.min_samples_leaf,
-                                            min_samples_split=self.min_samples_split,
-                                            random_state=self.random_state)
+        base_param_names = self._base_model_class._get_param_names()
```
vmarkovtsev: Is there a public function to get the names?

zurk: there is only a `get_params` one but it is not `@classmethod` so I can not use it here.

lookout/style/format/analyzer.py
zurk: I just this it is better to have the same order we have above: 
https://github.com/src-d/style-analyzer/pull/148/files/9c0c5615965b38a890760c7a2147482c1538cfe8#diff-b71e0e71a29c83c6fa2ee406a76c3baeR125

lookout/style/format/rules.py
vmarkovtsev: Will this work?
```python
self._base_model_class().get_params()
```

zurk: yup, I fix it and move to `base_model_name` property.

zurk:
it is tightly coupled with https://github.com/src-d/style-analyzer/pull/152 and now the part of it.

PR #140, opened 2018-09-21
Add generation module from #106 (4/6)

irinakhismatullina:
Module for candidate generation. Small examples of usage can be found in `test_generation.py`.

In a few words generation process works like this:
- Takes dataset of tokens to check with their context. 
- For every token from the dataset generates candidates  for correction (closest on edit distance to the tokens and most frequent & closest on embeddings cosine distance to the token and context).
- For every pair `(checked_token, candidate)` a list of features is generated. Edit distance, cosine distance between and vectors of checked token, candidate, context are taken.
- The dataset of candidates for every checked token and corresponding features is returned.

Ready to review and merge. Based on #109.

PR #151, opened 2018-09-24
Add corrector class (modelforge binding) from #106 (6/6)

irinakhismatullina:
Corrector class. 

- Accepts dataframes with columns "typo" for token to check, "identifier" for right correction (only for training), "token_split" for whole splitted identifier (not required, but used when present). Example values for columns `["typo", "identifier", "token_split"]`: `["tokem", "token", "get tokem count"]`.
- Generates candidates for correction with #140.
- Ranks candidates with #150.
- Returns candidates for correction. Suggestions filtering is also possible.

Ready to review and merge.


vmarkovtsev:
The code needs serious rewriting but let's do it after I merge all the pending PRs.

Issue #145, opened 2018-09-24
Feature extraction parameters should be saved in modelforge models.


m09:
At this point, the modelforge model contains, in addition to the rules themselves, the parameters used during the training of the rules. This doesn't include the parameters used during feature extraction, which are the only one needed to properly apply the rules to new data.

m09:
The saved parameters should also be used instead of the config ones in the `analyze` method of `analyzer.py`.

zurk:
I think it is also useful to add config info for unavoidable future model analysis.

m09:
yes, just saving the complete config seems reasonable to me.

vmarkovtsev:
@zurk Please post the PR here.

m09:
Fixed by #152.

PR #152, opened 2018-09-24
Add Rules config to FormatModel

zurk:
related issue: https://github.com/src-d/style-analyzer/issues/145

Also, I rename `origin` in `Rules` class to`origin_params` to make it more understandable and make a new name consistent with the name inside `FormatModel`.


lookout/style/format/model.py
```diff
@@ -21,9 +21,16 @@ class FormatModel(AnalyzerModel):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
         self._rules_by_lang = {}  # type: Dict[str, Rules]
+        self._config = {}  # type: Dict[str, Any]
+
+    def construct(self, analyzer: Type["Analyzer"], ptr: ReferencePointer, config: Dict[str, Any]
+                  ) -> "FormatModel":
+        super().construct(analyzer, ptr)
+        self._config = config
```
vmarkovtsev: The config should be different per language => we need to include it inside Rules.
In other words: Rules instance must contain all the configuration and there is no global one. Things like line filtering length are the parts of Rules config.

BTW - we should fix train() because currently it does not distinguish between languages.

zurk: also do not get about `train`.

zurk:
not outdated discussion here: https://github.com/src-d/style-analyzer/pull/152#discussion_r219829272

lookout/style/format/model.py
vmarkovtsev: `train()` has the same config for all the languages. Just checked the code - still does. It shouldn't.

vmarkovtsev: I like this config thing @m09 proposes.

m09:
@zurk I think it would be more straightforward to keep the config related code minimal and just have the different types of config as nested config. I find the current stuff with `origin`, `rest` and others a confusing. Ie:

```
config = {
    "feature_extraction": {
        "siblings_window": 5,
        …
    },
    "base_model": {
        …
    }
}

**Comment by zurk** [comment]
ok, I agree I was not sure about this idea. but if we do not want to split the config I should change Rules Model someway. Or we can keep it not in `Rules` class at all because all this code about `rest` required just to pass all arguments we do not need for training. We can create another `dict` in a modelforge model like `_config_by_lang`. But something gives me the cue that @vmarkovtsev against this solution.

@vmarkovtsev, @m09  WDYT?

**Comment by m09** [comment]
I would not be shocked if we passed the entire config to `Rules` (for me it's not a code smell at all to have a model and its parameters tightly coupled, quite the contrary). It's an integral part of the model (either to use the model during prediction or to know what the model is).

**Comment by vmarkovtsev** [comment]
Yes this is my vision too - same as @m09

**Comment by zurk** [comment]
ok, got it, guys!

**Comment by zurk** [comment]
Ok, I hope I did not do too deep and did not touch any interfaces I should not. 
I case I did tell me, guys.

File: lookout/core/analyzer.py [hunk 2]
```diff
@@ -68,7 +68,7 @@ class Analyzer:
     model_type = None  # type: Type[AnalyzerModel]
     name = None  # type: str
 
-    def __init__(self, model: AnalyzerModel, url: str, config: Mapping[str, Any]):
+    def __init__(self, model: AnalyzerModel, url: str, config: Mapping[str, Mapping[str, Any]]):
```
zurk: One question I have, how can we be sure that it is the same config we use for training?

zurk: ok, not the same, but all training related stuff is the same.

m09: we're not sure, that's why we have to save the config in modelforge for the parts that need to be the same for the model to work.

m09: the type should stay `Mapping[str, Any]` even though in our case it is more structured. It is explicit in the API that this parameter is free-form.

lookout/core/analyzer.py
```diff
@@ -97,8 +97,8 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
         raise NotImplementedError
 
     @classmethod
-    def train(cls, ptr: ReferencePointer, config: Dict[str, Any], data_request_stub: DataStub,
-              **data) -> AnalyzerModel:
+    def train(cls, ptr: ReferencePointer, config: Mapping[str, Mapping[str, Any]],
```
m09: Same as above. Should stay Mapping[str, Any]

lookout/style/format/analyzer.py
```diff
@@ -19,16 +19,25 @@
...
     name = "style.format.analyzer.FormatAnalyzer"
     version = "1"
     description = "Source code formatting: whitespace, new lines, quotes, braces."
 
-    def __init__(self, model: AnalyzerModel, url: str, config: Mapping[str, Any]) -> None:
+    def __init__(self, model: AnalyzerModel, url: str, config: Mapping[str, Mapping[str, Any]]
+                 ) -> None:
         super().__init__(model, url, config)
-        self.config = self._load_config(self.config)
+        self._config_generator = self._load_config(self.config)
```
m09: I don't think a generator is necessary. Just merging the default config with the given one and putting it in self.config for later use in ChainMap should be enough.

lookout/style/format/analyzer.py
```diff
@@ -62,10 +72,7 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
                     lines = None
                 else:
                     lines = [find_new_lines(prev_file, file)]
-                X, y, vnodes = FeatureExtractor(language=lang,
-                                                siblings_window=self.config["siblings_window"],
-                                                parents_depth=self.config["parents_depth"],
-                                                debug_parsing=self.config["debug_parsing"]) \
+                X, y, vnodes = FeatureExtractor(language=lang, **config["feature_extractor"]) \
```
m09: Since we're in the analyze method, the parameters for feature extraction should come from the model, not from the config.

lookout/style/format/analyzer.py
```diff
@@ -121,14 +129,7 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_request_st
...
             bscv = BayesSearchCV(
-                TrainableRules(
-                    prune_branches_algorithms=config["prune_branches_algorithms"],
-                    prune_attributes=config["prune_attributes"],
-                    top_down_greedy_budget=config["top_down_greedy_budget"],
-                    uncertain_attributes=config["uncertain_attributes"],
-                    prune_dataset_ratio=config["prune_dataset_ratio"],
-                    n_estimators=config["n_estimators"],
-                    random_state=config["random_state"]),
+                TrainableRules(config=config),
```
m09: This is one option (that we discussed yesterday and I'm ok with it). However I think it's silghtly better to require the config in the `FormatModel`'s `__setitem__`.

lookout/style/format/rules.py
```diff
@@ -108,8 +108,8 @@ def rules(self) -> List[Rule]:
         return self._rules
 
     @property
-    def origin(self) -> Mapping[str, Any]:
-        return self._origin
+    def config(self) -> Mapping[str, Any]:
```
m09: the property doesn't seem necessary

lookout/style/format/rules.py
```diff
@@ -589,3 +590,21 @@ def _sanitize_params(params: Dict[str, Any]) -> Dict[str, Any]:
                 v = list(v)
             sanitized[k] = v
         return sanitized
+
+    def _apply_config(self, config: Optional[Dict[str, Any]] = None) -> Dict[str, Any]:
```
m09: if we use the `__setitem__` method to get the config to modelforge, we can forgo this method and directly **config[key] they relevant key to initialize TrainableRules

lookout/core/analyzer.py
zurk: oh, sure!

lookout/style/format/analyzer.py
zurk: yes, you are right. I just think that in case we have a lot of languages no need to create all configs to analyze only one or two, which we have in the repo. Also, I do not want to make train function more complicated (even if it is just `ChainMap `). So I can create some object and make `config[language]` do all work in `train()`.
Is it ok?

lookout/style/format/analyzer.py
zurk: ok, so we have it, And everything is fine, nothing should be changed? did I get you right?

lookout/style/format/rules.py
zurk: yes, but I prefer to keep it because `_config` should be read-only. The property implements exactly this. Also before we have `origin` with the same meaning.

lookout/style/format/analyzer.py
zurk: so 
`TrainableRules(**config["trainable_rules"], config=config),`

lookout/style/format/rules.py
zurk: like this: https://github.com/src-d/style-analyzer/pull/152#discussion_r220448150

lookout/style/format/analyzer.py
m09: No sorry I meant only `TrainableRules(**config["trainable_rules"])` and then `model[language] = rules, config`. I'll PR feature selection this morning in which I have something like that (it's very similar to this PR in some sense so it'd make sense to coordinate reviews on both).

lookout/style/format/rules.py
m09: why not but we have plenty of readonly attributes in the codebase that are not `@property`'d. We should at least harmonize at some point.

lookout/style/format/analyzer.py
```diff
@@ -208,10 +230,10 @@ def _filter_files(cls, files: Iterable[File], line_length_limit: int) -> Iterabl
         :return: Files filtered.
         """
         excluded = total = 0
-        for file in files:
+        for filename, file in files.items():
             if len(max(file.content.splitlines(), key=len, default=b"")) <= line_length_limit:
                 total += 1
-                yield file
+                yield filename, file
```
vmarkovtsev: -1. `file` already has the path property. Please remove.

lookout/style/format/rules.py
```diff
@@ -49,18 +49,18 @@ class Rules:
 
     _log = logging.getLogger("Rules")
 
-    def __init__(self, rules: List[Rule], origin: Mapping[str, Any]):
+    def __init__(self, rules: List[Rule], config: Mapping[str, Any]):
```
vmarkovtsev: `origin` is more expressive here because we do not parameterize anything here, just include the origin-al configuration.
It's like naming a counter "integer": yep, it's an integer but no use of that name.

lookout/style/format/rules.py
vmarkovtsev: +1 for leaving the property and adding other readonly properties.

lookout/style/format/analyzer.py
zurk: ok, just wanted to play with generators ;)

lookout/style/format/analyzer.py
zurk: Ok, it is not a problem and I like it more, but if I remember correctly @vmarkovtsev want to store config inside rules: https://github.com/src-d/style-analyzer/blob/cfb9c2e9330bde757181537592849181420202f0/lookout/style/format/rules.py#L278

 @vmarkovtsev do you mind if we move it from Rules to just Modelforge model?

lookout/style/format/rules.py
zurk: can I use `origin_config`? because IMHO `origin` also not understandable.

lookout/style/format/rules.py
m09: AFAIK all params of trainable rules are read only for example. Same for many other params in many other places

zurk: oh, of course I mean this class only. But I keep it as an issue. Better to have separate PR for that.

lookout/style/format/analyzer.py
vmarkovtsev: So we shouldn't have any configs in the model. Our model is a dictionary of Rules per language. Each Rules must provide enough information to reproduce it bit-to-bit in form of a configuration dictionary. Model is super dumb and must continue to be dumb.

Some configuration bits will be the same across Rules instances and that it normal.

zurk: ok, I will put it to model doc

zurk:
WIP.

lookout/core/analyzer.py
```diff
@@ -68,18 +68,16 @@ class Analyzer:
     model_type = None  # type: Type[AnalyzerModel]
     name = None  # type: str
 
-    def __init__(self, model: AnalyzerModel, url: str, config: Mapping[str, Any]):
+    def __init__(self, model: AnalyzerModel, url: str):
```
m09: This breaks the API. We should still have config here even though we don't use it yet in this precise analyzer (in function name analyzer this is the only config we should use for example).

lookout/style/format/analyzer.py
```diff
@@ -26,9 +26,9 @@ class FormatAnalyzer(Analyzer):
     version = "1"
     description = "Source code formatting: whitespace, new lines, quotes, braces."
 
-    def __init__(self, model: AnalyzerModel, url: str, config: Mapping[str, Any]) -> None:
-        super().__init__(model, url, config)
-        self.config = self._load_config(self.config)
+    def __init__(self, model: FormatModel, url: str) -> None:
```
m09: cf other comment, we should not remove config here.

zurk:
but I do not see your review...

m09:
yeah I removed it since you added WIP. I let the comment because you probably received mails.

zurk:
ok, I have a question about configs.
as I remember we should keep config in `__init__` since it is different from config we use to train the model. So I think it is a good idea to create `_train_config` where we store config restored from `FormatModel` and `_config` where we store independent config for `analyze` part.

@m09 WDYT?

zurk:
ready.

lookout/style/format/analyzer.py
```diff
@@ -44,44 +43,40 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
         """
         comments = []
         changes = list(data["changes"])
-        base_files = self._files_by_language(self._filter_files((c.base for c in changes),
-                                                                self.config["line_length_limit"]))
-        head_files = self._files_by_language(self._filter_files((c.head for c in changes),
-                                                                self.config["line_length_limit"]))
+        base_files = self._files_by_language((c.base for c in changes))
+        head_files = self._files_by_language((c.head for c in changes))
```
m09: Please rename `base_files_by_lang` and `head_files_by_lang`

lookout/style/format/analyzer.py
```diff
@@ -44,44 +43,40 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
         for lang, lang_head_files in head_files.items():
```
m09: Please rename lang_head_files to head_files

lookout/style/format/analyzer.py
```diff
@@ -44,44 +43,40 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
-            try:
-                rules = self.model[lang]
-            except KeyError:
-                self.log.warning("skipped %d written in %s - model does not exist",
-                                 len(lang_head_files), lang)
+            if lang not in self.model:
+                self.log.warning("skipped %d written in %s. Rules for %s does not exist in model",
```
m09: do not exist in the model*

lookout/style/format/analyzer.py
```diff
@@ -44,44 +43,40 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
         """
         comments = []
         changes = list(data["changes"])
-        base_files = self._files_by_language(self._filter_files((c.base for c in changes),
-                                                                self.config["line_length_limit"]))
-        head_files = self._files_by_language(self._filter_files((c.head for c in changes),
-                                                                self.config["line_length_limit"]))
+        base_files = self._files_by_language((c.base for c in changes))
```
m09: no need for the inner parens when the generator is the only argument

lookout/style/format/analyzer.py
```diff
@@ -177,29 +162,43 @@ def _files_by_language(cls, files: Iterable[File]) -> Dict[str, Dict[str, File]]
...
-            "debug_parsing": False,
-            "line_length_limit": 500
+        def merge(default_config, config):
+            for key, value in config.items():
+                if isinstance(value, dict):
+                    merge(default_config.setdefault(key, {}), value)
+                else:
+                    default_config[key] = value
+
+        config_ = {
```
m09: Please rename `final_config`

lookout/style/format/analyzer.py
```diff
@@ -177,29 +162,43 @@ def _files_by_language(cls, files: Iterable[File]) -> Dict[str, Dict[str, File]]
...
-            "top_down_greedy_budget": [False, .5],
-            "prune_attributes": False,
-            "uncertain_attributes": True,
-            "prune_dataset_ratio": .2,
-            "n_estimators": 10,
-            "random_state": 42,
-            "n_iter": 5,
-            "debug_parsing": False,
-            "line_length_limit": 500
+        def merge(default_config, config):
```
m09: Please rename the function `recursive_update` and the arguments `mapping` and `other` to follow standard lib a bit more

lookout/style/format/rules.py
```diff
@@ -190,23 +190,25 @@ def __init__(self, base_model_name: str = "sklearn.tree.DecisionTreeClassifier",
...
-        :param prune_base_model: indicates whether to prune base_model via reduced error pruning \
-                                 algorithm. Available for DecisionTreeClassifier model only.
+        :param origin_config: all parameters that are used for the model training.
         """
         super().__init__()
+
         self.base_model_name = base_model_name
         self.prune_branches_algorithms = prune_branches_algorithms
-        self.top_down_greedy_budget = top_down_greedy_budget
+        self.top_down_greedy_budget = TopDownGreedyBudget(*top_down_greedy_budget)
```
m09: :+1:

zurk:
Now everything is addressed. 🎉

PR #156, opened 2018-09-26
Add feature selection

m09:
This touches quite some classes since we didn't include feature extraction as an integral part of the model yet.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -7,6 +7,7 @@ from typing import (Callable, Dict, Iterable, List, Mapping, NamedTuple, Optiona
 
 import bblfsh
 import numpy
+from sklearn.feature_selection import SelectKBest, VarianceThreshold
 
 from lookout.core.api.service_analyzer_pb2 import Comment
 from lookout.core.api.service_data_pb2 import File
@@ -170,7 +171,8 @@ class FeatureExtractor:
     _log = logging.getLogger("FeaturesExtractor")
 
     def __init__(self, language: str, *, siblings_window: int = 5, parents_depth: int = 2,
-                 debug_parsing: bool = False):
+                 select_features_number: Optional[int] = None, remove_empty_features: bool = True,
+                 selected_features: Optional[numpy.ndarray] = None, debug_parsing: bool = False):
         """
         Construct a `FeatureExtractor`.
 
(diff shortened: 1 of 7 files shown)


m09:
This also contains some form of #155 so I'll rebase when #155 is merged :)

EgorBu:
yes, it could be easier to merge #155 - it's much smaller)

lookout/style/format/features.py
```diff
@@ -414,24 +432,26 @@ def _add_noops(vnodes: Sequence[VirtualNode], path: str) -> List[VirtualNode]:
...
-    def _get_internal_type_index(self, vnode: VirtualNode) -> int:
-        role_index = -1
-        if vnode.node:
-            role = vnode.node.internal_type
-            if role in self.roles.INTERNAL_TYPES_INDEX:
-                role_index = self.roles.INTERNAL_TYPES_INDEX[role]
-        return role_index
+    def _get_internal_type_index(self, node: Union[bblfsh.Node, VirtualNode]) -> int:
+        internal_type_index = -1
+        bblfsh_node = node if isinstance(node, bblfsh.Node) else node.node
```
warenlg: I met an `AttributeError` here because in some cases node is `gopkg.in.bblfsh.sdk.v1.uast.generated_pb2.Node` class so `isinstance(node, bblfsh.Node)` retruns `False`

m09: @vmarkovtsev wasn't this fixed in https://github.com/bblfsh/client-python/pull/124?
@waren I'll change the test (use virtual node to test and default to bblfsh

m09:
@vmarkovtsev I didn't ask for your review since you were giving talks but if it's not too painful on your phone you can TAL :)

lookout/style/format/features.py
```diff
@@ -241,7 +245,7 @@ def _init_feature_names(self, feature_layout):
         self._feature2index = {name: i for i, name in enumerate(self.feature_names)}
 
     def extract_features(self, files: Iterable[File], lines: List[List[int]]=None
-                         ) -> Optional[Tuple[numpy.ndarray, numpy.ndarray, List[VirtualNode]]]:
+                         ) -> Tuple[numpy.ndarray, numpy.ndarray, List[VirtualNode]]:
```
zurk: But I see we can return `None` here: https://github.com/src-d/style-analyzer/pull/156/files#diff-f778fb4562b4ded91223ccd22dd95b08L279. So it is optional or 
we can fix return value to `return numpy.empty(shape=(0, 1)), numpy.empty(shape=(0, self.count_features(FeatureType.all))), []` because we do `a,b,c = fe.extract_features(...)` and it can failed.

m09: No no you're right it's an oversight during refactoring, thanks for catching it :+1:

Issue #157, opened 2018-09-26
Style analyser failed to analyse with relation "models" does not exist


zurk:
```
➜  style-analyzer git:(master) python3 -m lookout run lookout.style.format -s 0.0.0.0:2000 --db postgres://postgres:postgres@0.0.0.0:5432/lookout --fs /tmp --log-level DEBUG
INFO:2d13:run:Created SQLAlchemyModelRepository(db=postgres://postgres:postgres@0.0.0.0:5432/lookout, fs=/tmp)
INFO:2d13:run:Created DataService(0.0.0.0:10301)
/usr/local/lib/python3.7/site-packages/sklearn/ensemble/weight_boosting.py:29: DeprecationWarning: numpy.core.umath_tests is an internal NumPy module and should not be imported. It will be removed in a future NumPy release.
  from numpy.core.umath_tests import inner1d
INFO:2d13:run:Created AnalyzerManager(style.format.analyzer.FormatAnalyzer/1)
INFO:2d13:run:Created EventListener(0.0.0.0:2000, 1 workers)
INFO:2d13:run:Listening 0.0.0.0:2000
INFO:51f5:EventListener:new ReviewEvent
DEBUG:AnalyzerManager:no config was provided for style.format.analyzer.FormatAnalyzer
ERROR:51f5:EventListener:FAIL ?
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1193, in _execute_context
    context)
  File "/usr/local/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 509, in do_execute
    cursor.execute(statement, parameters)
psycopg2.ProgrammingError: relation "models" does not exist
LINE 2: FROM models
             ^


The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/k/sourced/workdir2/style-analyzer/lookout/core/event_listener.py", line 161, in wrapped_catch_them_all
    return func(self, request, context)
  File "/Users/k/sourced/workdir2/style-analyzer/lookout/core/event_listener.py", line 182, in wrapped_handle
    return getattr(self.handlers, method_name)(request)
  File "/Users/k/sourced/workdir2/style-analyzer/lookout/core/manager.py", line 61, in process_review_event
    self._model_id(analyzer), analyzer.model_type, base_ptr.url)
[...]

PR #158, opened 2018-09-26
Fix "relation "models" does not exist"

zurk:
Fixes https://github.com/src-d/style-analyzer/issues/157


zurk:
code coverage only. @vmarkovtsev can you review?

PR #164, opened 2018-09-26
Downgrade sklearn to remove incompatibility with skopt

m09:
Fixes travis.


zurk:
oh, that what it was. this issue just disappeared for me :)

PR #153, opened 2018-09-25
Reduce memory usage

m09:

PR #165, opened 2018-09-27
Fix review & config

EgorBu:
* Fix review for broken files
* comment for unparsable files
* config for `n_jobs` - could be useful for big repos to avoid OOM errors

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -59,9 +59,18 @@ class FormatAnalyzer(Analyzer):
                     lines = None
                 else:
                     lines = [find_new_lines(prev_file, file)]
-                X, y, vnodes = FeatureExtractor(language=lang,
-                                                **rules.origin_config["feature_extractor"]) \
+                res = FeatureExtractor(language=lang,
+                                       **rules.origin_config["feature_extractor"]) \
                     .extract_features([file], lines)
+                if res is None:
+                    comment = Comment()
+                    comment.file = file.path
+                    comment.confidence = 1.
+                    comment.line = 1
+                    comment.text = "Failed to parse this file"
+                    continue
+                else:
+                    X, y, vnodes = res
                 self.log.debug("predicting values for %d samples", len(y))
                 y_pred, winners = rules.predict(X, True)
                 assert len(y) == len(y_pred)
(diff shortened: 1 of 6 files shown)


lookout/style/format/analyzer.py
```diff
@@ -135,7 +142,7 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_request_st
                  "max_features": Categorical([None, "auto"]),
                  "min_samples_split": Integer(2, 20),
                  "min_samples_leaf": Integer(1, 20)},
-                n_jobs=-1,
+                n_jobs=config["n_jobs"],
```
m09: +1
However I fear that when we need it (ie big repos) it will become so slow that it'll also be unbearable. We could probably experiment with subsambling if we want to find another way to mitigate both memory and speed issues.

EgorBu:
when we parse broken file - feature extractor will try to extract features from it.
It's not so important for training - we extract features from many files, but important for review - we extract features for each specific file separately. Because of this [here](https://github.com/src-d/style-analyzer/blob/688d3468a0da4b4178f49e44c43ac5b222d460ec/lookout/style/format/features.py#L280) we will have an error  because numpy array is not possible create from empty list - so I added check that we have some features and return triple None in case of failure (triple - because it should return 3 items in result)

lookout/style/format/features.py
```diff
@@ -273,6 +273,10 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None
             labels.append([vnode.y for vnode in vnodes if vnode.y is not None and
                            (vnode.start.line in file_lines if file_lines is not None else True)])
 
+        if not labels:
+            # nothing was extracted
+            return None, None, None
```
m09: I think it might be better to return only one `None` and have the type changed to `Optional[Tuple[…]]`. We'll have to check for `None`-s now and won't ever check on the 3 values independently so it might be more consistent with our usage.

lookout/style/format/analyzer.py
EgorBu: at least it could solve the problem for local experiments when RAM is quite limited + this argument allows to reduce number of iterations https://github.com/src-d/style-analyzer/blob/688d3468a0da4b4178f49e44c43ac5b222d460ec/lookout/style/format/analyzer.py#L146

m09: true. Btw you can rebase #153 if you want to reduce memory impact, it helps a lot already :)

lookout/style/format/features.py
EgorBu: yes, it's a good option because tuple of 3 None's is not the most obvious thing you expect to be returned.

PR #167, opened 2018-09-27
replace < and > by ` and ` in comments

zurk:
because now it is invisible in comment:
```
format: replace <newline> with <space> at column 5
```
 converts to 
```
format: replace with at column 5
```


EgorBu:
why not inline code?
```comment.text.replace("<", "`<").replace(">", ">`")``` - so the name of class will be the same.

zurk:
anyway, it is not the final option, because it is ugly :)
Now we need to have at least something. If most of us likes `<>` more I can change it.

m09:
It's definitely not final and so we should not care too much about the very precise details of the message. This is a just a bugfix to make the class names appear after all.

PR #168, opened 2018-09-27
Identifiers typos analyzer

irinakhismatullina:
**IdTyposAnalyzer workflow**:
- Extracts new identifiers for the changed file.
- Passes identifiers to IdTyposModel, the model returns suggestions for spelling correction.
- Composes comments based on this suggestions and sends them to lookout.

**IdTyposModel workflow**:
- Get list of identifiers.
- Split every identifier (using sourced.ml TokenParser, should be changed to the cool DL id splitter)
- Create dataframe with info about every token inside every identifier to check them. Every token is considered a possible typo.
- Send them to the pretrained TyposCorrector model (it is an ItTyposModel class attribute).
- Get corrections suggestions from the corrector. Filter them based on the confidence threshold and number of candidates to return.
- Return filtered corrections suggestions.

Ready for review. Most probably needs refactoring. Based on #151.


vmarkovtsev:
Had to skip the tests in IdTyposAnalyzer because obviously it was not finished: `train()` was a No-op.

vmarkovtsev:
Entering the huge rewriting stage.

PR #174, opened 2018-09-27
Sequence matching research

EgorBu:
Sequence matching research - script for visualization.
Depends on https://github.com/src-d/style-analyzer/pull/165


m09:
This file needs some updates to fit to current pipeline but it's in research so I'll merge as is and we'll update if we use it and want to merge it to main viz code :)

Issue #180, opened 2018-09-28
FeaturesExtractor:could not parse file common/utils/encode-decode.js with error 'missed some imaginary tokens:


smola:
I get this warning:

```
WARNING:4af0:FeaturesExtractor:could not parse file common/utils/encode-decode.js with error 'missed some imaginary tokens: "
    /<form[^>]*>/,
    "', skipping
```

The problem seems to be here: https://github.com/jlouiss/freeCodeCamp/blob/ff07a3637863ed15439acab642818d6e368d3c08/common/utils/encode-decode.js#L23

zurk:
ok, I found the reason. https://github.com/bblfsh/javascript-driver/issues/37, for now, I add a workaround. Sorry, @m09 for stealing your issue :)

m09:
This is normally handled by
```
            if (node.token or node.start_position and node.end_position
                    and node.start_position != node.end_position and not node.children):
                node_tokens.append(node)
```
This case fails for some other reason if I didn't miss anything.

zurk:
the problem is in a comment here. Because the comment is a child of regexp node:
for this code
```js
x = // comment
  /<regexp>/;
```

UAST gives:
```
#           Token       Internal Role         Roles Tree                                                    
                                                                                                            
(1, 1, 12)  ||          File                  FILE                                                          
(1, 1, 12)  ||          Program               ┣ MODULE                                                      
(1, 1, 12)  ||          ExpressionStatement   ┃ ┣ STATEMENT                                                 
(1, 1, 11)  ||          AssignmentExpression  ┃ ┃ ┣ EXPRESSION, ASSIGNMENT, OPERATOR, BINARY                
(1, 1, 2)   |x|         Identifier            ┃ ┃ ┃ ┣ EXPRESSION, IDENTIFIER, ASSIGNMENT, BINARY, LEFT      
(2, 1, 11)  ||          RegExpLiteral         ┃ ┃ ┃ ┣ EXPRESSION, LITERAL, REGEXP, ASSIGNMENT, BINARY, RIGHT
(1, 5, 15)  | comment|  CommentLine           ┃ ┗ ┗ ┗ ┗ COMMENT                                             
(1, 5, 15)  | comment|  CommentLine           ┗ COMMENT  
```

PR #181, opened 2018-09-28
Add new test for _parse_file

zurk:
`"".join(n.value for n in nodes)` where `nodes, _ = self.extractor._parse_file(...)` produce bad result.
here is the diff from added test:

```javascript
- // from https://github.com/freeCodeCamp/freeCodeCamp/blob/master/server/component-passport.js// from https://github.com/freeCodeCamp/freeCodeCamp/blob/master/server/component-passport.js
+ // from https://github.com/freeCodeCamp/freeCodeCamp/blob/master/server/component-passport.js
  import passport from 'passport';
- import { PassportConfiguratorPassportConfigurator } from
?                              --------------------
+ import { PassportConfigurator } from
   '@freecodecamp/loopback-component-passport';
  import passportProviders from './passport-providers';
  import url from 'url';
  import jwt from 'jsonwebtoken';
  import dedent from 'dedent';
  
  const passportOptions={
    emailOptional: true,
    profileToUser: null
  };
  
  const fields = {
    progressTimestamps: false,
  };
  
  function getCompletedCertCount(user) {
    return [
      'isApisMicroservicesCert',
      'is2018DataVisCert',
      'isFrontEndLibsCert',
      'isInfosecQaCert',
      'isJsAlgoDataStructCert',
      'isRespWebDesignCert'
    ].reduce((sum, key) => user[key] ? sum + 1 : sum, 0);
  }
  
  function getLegacyCertCount(user) {
    return [
      'isFrontEndCert',
      'isBackEndCert',
      'isDataVisCert'
    ].reduce((sum,key) => user[key] ? sum + 1 : sum, 0);
  }
  
  PassportConfigurator.prototype.init = function passportInit(noSession) {
    this.app.middleware('session:after', passport.initialize());
  
    if (noSession) {
      return;
    }
  
    this.app.middleware('session:after',passport.session());
  
-   // Serialization and deserialization is only required if passport session is// Serialization and deserialization is only required if passport session is// Serialization and deserialization is only required if passport session is
-   // enabled// enabled// enabled
+   // Serialization and deserialization is only required if passport session is
+   // enabled
  
    passport.serializeUser((user, done) =>
    {
      done(null, user.id);
    });
  
    passport.deserializeUser((id, done) => {
  
-     this.userModel.findById(id, { fieldsfields }, (err, user) => {
?                                         ------
+     this.userModel.findById(id, { fields }, (err, user) => {
        if (err || !user) {
          return done(err, user);
        }
  
        return this.app.dataSources.db.connector
          .collection('user')
          .aggregate([
            { $match: { _id: user.id } },
            { $project: { points: { $size: '$progressTimestamps' } } }
-         ], function(err, [{ pointspoints = 1 } = {}]) {
?                                   ------
+         ], function(err, [{ points = 1 } = {}]) {
            if (err) { return done(err); }
            user.points = points;
            let completedChallengeCount = 0;
            let completedProjectCount = 0;
            if ('completedChallenges' in user) {
              completedChallengeCount = user.completedChallenges.length;
              user.completedChallenges.forEach(item => {
                if (
                  'challengeType' in item &&
                  (item.challengeType === 3 || item.challengeType === 4)
                ) {
                  completedProjectCount++;
                }
              });
            }
            user.completedChallengeCount = completedChallengeCount;
            user.completedProjectCount = completedProjectCount;
            user.completedCertCount = getCompletedCertCount(user);
            user.completedLegacyCertCount = getLegacyCertCount(user);
            user.completedChallenges = [];
            return done(null, user);
          });
      });
    });
  };
  
  export default function setupPassport(app) {
    const configurator = new PassportConfigurator(app);
  
    configurator.setupModels({
      userModel: app.models.user,
      userIdentityModel: app.models.userIdentity,
      userCredentialModel: app.models.userCredential
    });
  
    configurator.init();
  
    Object.keys(passportProviders).map(function(strategy) {
      let config = passportProviders[strategy];
      config.session = config.session !== false;
  
-     // https://stackoverflow.com/q/37430452// https://stackoverflow.com/q/37430452// https://stackoverflow.com/q/37430452
+     // https://stackoverflow.com/q/37430452
      let successRedirect = (req) => {
        if (!!req && req.session && req.session.returnTo) {
          delete req.session.returnTo;
          return'/';}
        return config.successRedirect || '';
      };
  
      config.customCallback = !config.useCustomCallback
        ? null
        : (req, res, next) => {
  
          passport.authenticate(
            strategy,
            { session: false },
            (err, user, userInfo) => {
  
              if (err) {
                return next(err);
              }
  
              if (!user || !userInfo) {
                return res.redirect(config.failureRedirect);
              }
              let redirect = url.parse(successRedirect(req), true);
  
              delete redirect.search;
  
-             const {accessTokenaccessToken} = userInfo;
?                               -----------
+             const {accessToken} = userInfo;
-             const { providerprovider } = config;
?                             --------
+             const { provider } = config;
              if (accessToken && accessToken.id) {
                if (provider === 'auth0') {
                  req.flash(
                    'success',
                    dedent`
                      Success! You have signed in to your account. Happy Coding!
                    `
                  );
                } else if (user.email) {
                  req.flash(
                    'info',
                    dedent`
    We are moving away from social authentication for privacy reasons. Next time
    we recommend using your email address: ${user.email} to sign in instead.
                    `
                  );
                }
                const cookieConfig = {
                  signed: !!req.signedCookies,
                  maxAge:accessToken.ttl,
                  domain: process.env.COOKIE_DOMAIN || 'localhost',
                };
-               const jwtAccess = jwt.sign({accessTokenaccessToken}, process.env.JWT_SECRET);
?                                                      -----------
+               const jwtAccess = jwt.sign({accessToken}, process.env.JWT_SECRET);
                res.cookie('jwt_access_token', jwtAccess, cookieConfig);
                res.cookie('access_token', accessToken.id, cookieConfig);
                res.cookie('userId', accessToken.userId, cookieConfig);
  
                req.login(user);
              }
  
              redirect = url.format(redirect);
              return res.redirect(redirect);
            }
          )(req, res, next);
      };
  
      configurator.configureProvider(
        strategy,
        {
          ... config,
          ...passportOptions
        }
      );
    });}
```


zurk:
realted to https://github.com/bblfsh/javascript-driver/issues/36

lookout/style/format/tests/test_features.py
```diff
@@ -25,6 +25,15 @@ def setUpClass(cls):
             cls.uast = bblfsh.Node.FromString(fin.read())
         cls.extractor = FeatureExtractor("javascript", parents_depth=2, siblings_window=5)
 
+    def test_parse_file2(self):
```
m09: Please rename to something more informative like test_parse_file_on_comments if comments are causing a problem for example.

lookout/style/format/tests/test_features.py
```diff
@@ -25,6 +25,15 @@ def setUpClass(cls):
...
+        test_js_code_filepath = Path(__file__).parent / "for_parse_test.js.xz"
+        with lzma.open(str(test_js_code_filepath), mode="rt") as f:
+            code = f.read()
+        uast = bblfsh.BblfshClient("0.0.0.0:9432").parse(
+            filename="", language="javascript", contents=code.encode()).uast
```
m09: no need to specify filename when contents is set. It could also be good to just bypass babelfish and just use a saved uast as we do in the other tests. Up to you :)

zurk: nope. here the header of parse function: `def parse(self, filename, language=None, contents=None, timeout=None):` also the file name goes to bblfsh logs.
When we update to the next version of a bblfsh driver, we have to update this files.
I prefer to have it like that.

lookout/style/format/tests/test_features.py
zurk: mmm, it is not only for comments. It is for random stuff. I rename it to `test_parse_file_exact_match`

lookout/style/format/tests/test_features.py
m09: Fair enough, even though we will have to update all the other files in the folder anyway. And huh, this function signature needs refactoring, I'll open an issue.

zurk:
No, I check it and it does not fix #180

PR #186, opened 2018-09-29
Add debug log output of base_model_name feaure importance

zurk:
Format:
```
DEBUG:FormatAnalyzer:Feature importances from sklearn.ensemble.RandomForestClassifier:
	left_1_y_is_<space>           : 1.59749E-01
	left_1_reserved_is_throw      : 6.28934E-02
	left_1_reserved_is_new        : 6.16138E-02
	left_2_y_is_<+space>          : 6.01231E-02
	left_1_reserved_is_super      : 5.04645E-02
	left_1_y_is_<+space>          : 3.85661E-02
	left_1_reserved_is_string     : 3.77613E-02
	left_1_reserved_is_%=         : 3.71012E-02
	left_1_reserved_is_]          : 2.79625E-02
	left_1_role_is_expression     : 2.62532E-02
...
```

### DIFF
--- a/lookout/style/format/rules.py
+++ b/lookout/style/format/rules.py
@@ -220,6 +220,7 @@ class TrainableRules(BaseEstimator, ClassifierMixin):
         :param y: input labels - the same length as X.
         :return: self
         """
+        self._log.debug("fitting rules with params %s", self.get_params(False))
         models_params = {name: val for name, val in self.get_params().items()
                          if name in self._base_param_names}
         base_model = self._base_model_class(**models_params)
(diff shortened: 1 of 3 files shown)


lookout/style/format/rules.py
```diff
@@ -220,6 +220,7 @@ def fit(self, X: numpy.ndarray, y: numpy.ndarray) -> "TrainableRules":
         :param y: input labels - the same length as X.
         :return: self
         """
+        self._log.debug("TrainableRules.fit(...)")
```
m09: Can you please rewrite as `"fitting rules with config %s" % self.origin_config["trainable_rules"]` to be more consistent with our other logs and give a bit more info.

zurk: I think it is better to have something like
```
self._log.debug("fitting rules with params %s", self.get_params(False))
```
because we use only this params  from config in fit().

m09: I think both are pretty similar no? Since with `origin_config["trainable_rules"]` we only pick the section about trainable rules. But yeah both are fine for me.

zurk: mix up things, you are right.

zurk:
addressed.

PR #189, opened 2018-09-29
Add _load_analyze_config to Analyser

zurk:
Also, rename `_load_config` to `_load_train_config`
I need `_load_analyze_config` to have more verbose comments when I develop.


lookout/style/format/analyzer.py
```diff
@@ -172,7 +172,16 @@ def _files_by_language(cls, files: Iterable[File]) -> Dict[str, Dict[str, File]]
         return result
 
     @classmethod
-    def _load_config(cls, config: Mapping[str, Any]):
+    def _load_analyze_config(cls, config: Mapping[str, Any]) -> Mapping[str, Any]:
```
m09: please use the same mechanism as in train_config. Probably it's best to make recursive_update an external function to avoid duplicates.

zurk: yeah, it is not necessary right now so, I wanted to avoid it, but ok.

m09: I agree but it's not long to do and it's good to have consistency.

lookout/core/analyzer.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, model: AnalyzerModel, url: str, config: Mapping[str, Any]):
         """
         self.model = model
         self.url = url
-        self.config = config
+        self.config = self._load_analyze_config(config)
```
m09: this should not be in core but in format analyzer

lookout/core/analyzer.py
```diff
@@ -112,6 +112,16 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any],
         """
         raise NotImplementedError
 
+    @classmethod
```
m09: this is not part of the API so we should just implement it in format analyzer and not touch the core package

lookout/core/tests/test_manager.py
```diff
@@ -40,6 +40,9 @@ def train(cls, ptr: ReferencePointer, config: dict, data_request_stub: DataStub,
         cls.stub = data_request_stub
         return FakeModel()
 
+    def _load_analyze_config(cls, config):
```
m09: can be removed after the package change

zurk:
addressed.

PR #190, opened 2018-09-29
Feature/informative comment

zurk:
This commit contains several things. It is hard to split them into several because they are not going to work without each other. 
The best thing: this commit makes style analyzer produce meaningful comments. 

Changes list: 
1. New feature matrix. I add only several features and only for left tokens: one-hot-encoded (OHE) for reserved tokens, OHE for `y` (target) and 2 roles indicators: for `EXPRESSION` and `IDENTIFIER` roles. No right side, no parents, no self, no positions at all. Performance on reproduction task is `0.8484`. It is smaller than what we had, but comments are much better. You can check out usage example here: zurk/js_test_lookout#3 
2. `humanize()` for `Rules` and `RuleAttribute` classes. Now we can convert unreadable rules into something useful:
```
    left_1_reserved in |(˙,˙;˙None˙{˙}|
    left_3_role in |expression|
    left_5_role in |expression|
```
`˙` is just separator. This part is connected to `1.` because of the special feature name structure. I think it can be avoided in the future. But now it is hard to do.
3. Output comment was improved a lot. I include all info that can help us better understand what is going on.

@m09 I know you work on feature matrix configurator. I think after you PR it I can reimplement this PR using new tooling. It will be cool if somehow I will be able to determine features from the same group (like OHE features) to humanize our rules without add-hoc code, like now.

So this PR is more for informational purposes now.


m09:
It's good to have checked that without leakage the comments improve but I think it's important not to throw out the baby with the bathwater. For example, removing only right siblings reduces the precision to 95% but probably still gets rid of leakage. Pending confirmation (ie checking if comments are sensible), we'll make around 3 times less false positives than with a leakage free but 84% precision system. Anyway it's a great first step :tada:

zurk:
yes, you are absolutely right. This is the first step, I just wanted to start from the opposite situation we have.
We were in a situation where nothing works, but we have a lot of features, so we should exclude one by one and track improvements. it hard to do.
So in my case, I start with almost zero features. I iteratively add feature by feature and improve quality. It is just easier to analyze. And, yeah, no data leakage guaranteed. :)

In tomorrow plans, I want to try adding more features, I think for parents and the right tokens. I think it should fix some of the strange false positives I saw. I already tried to use line and column information and it gives me bad results: rules are less interpretable and strange false positives (like <space+> something where you do not expect it.)

Also if have time experiment with NOOP. without any code change, just try to turn it on back. 

In global plans, we really need to have the dataset. Because it is impossible to measure anything right now. I see here two steps:
1. Each ML engineer creates 10 files, where some style is used. then creates 10 style mistakes in them. So 100 samples from each: 500 samples -- not so bad for the beginning. 
2. I hope we will use style analyzer internally and should collect user feedback: it helps us to control our false positives.

m09:
We should take care not to duplicate the work Waren has been doing though: he already has a dataset with mistakes and has started to work on two metrics to get direct performance from a system.

lookout/style/format/analyzer.py
```diff
@@ -28,6 +28,7 @@ class FormatAnalyzer(Analyzer):
 
     def __init__(self, model: FormatModel, url: str, config: Mapping[str, Any]) -> None:
         super().__init__(model, url, config)
+        self.config = self._load_analyze_config(self.config)
```
vmarkovtsev: We should rename this to `_load_inference_config`

lookout/style/format/analyzer.py
```diff
@@ -41,6 +42,22 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
         :param data: Contains "changes" - the list of changes in the pointed state.
         :return: List of comments.
         """
+        def one_line_nodes(y, y_pred, vnodes, rule_winners):
```
vmarkovtsev: There should be a verb which describes this function. What does it do?

lookout/style/format/analyzer.py
```diff
@@ -67,19 +83,35 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
-                        comments.append(comment)
+                if self.config["debug"]:
+                    code_lines = file.content.decode("utf-8", "replace").splitlines()
+                for line_nodes in one_line_nodes(y, y_pred, vnodes, rule_winners):
+                    code_line_number = line_nodes[0][2].start.line - 1
+                    code_text = (
+                        "" if not self.config["debug"] else
+                        "```%s\n%s\n```\n\n" % (lang, code_lines[code_line_number]))
+                    vnodes_comments = ["%s\n%s\n" % (
+                        vnode.to_comment(y_predi),
```
vmarkovtsev: We should rename `to_comment` because the new function does not really produce a comment - it produces a string which we use as a comment, but don't have to.

vmarkovtsev: I would suggest to remove it completely and define one here.

vmarkovtsev: In the future we will try to make it sexy and move all the code here inside the Rule

lookout/style/format/analyzer.py
```diff
@@ -172,14 +204,21 @@ def _files_by_language(cls, files: Iterable[File]) -> Dict[str, Dict[str, File]]
...
+        """
+        Merges config for analyze call with default config values stored inside this function.
+
+        :param config: User-defined config.
+        :return: Full config.
+        """
+        final_config = {
+            "debug": False,
+        }
+        FormatAnalyzer.recursive_update(final_config, config)
```
vmarkovtsev: ~Wow this name sucks so much!~

The best here would be to move this `recursive_update` out of the class and name it `merge_dicts`

lookout/style/format/analyzer.py
```diff
@@ -230,3 +269,11 @@ def _filter_files(cls, files: Dict[str, File], line_length_limit: int
         if excluded > 0:
             cls.log.debug("excluded %d/%d files by max line length %d",
                           excluded, total, line_length_limit)
+
+    @staticmethod
+    def recursive_update(mapping, other):
+        for key, value in other.items():
+            if isinstance(value, dict):
+                FormatAnalyzer.recursive_update(mapping.setdefault(key, {}), value)
```
vmarkovtsev: Please change the implementation from recursive to non-recursive. It ts very easy.

lookout/style/format/features.py
```diff
@@ -199,21 +196,17 @@ def __init__(self, language: str, *, siblings_window: int = 5, parents_depth: in
             # It's normal for some languages not to have a token_unwrappers module.
             self.token_unwrappers = {}
 
-        # Order is important and should be consistent with _inplace_write_vnode_features function
-        # where features generation happens.
+        def prefix(prefix, l):
```
vmarkovtsev: prefix_sequence_elements

Also the types are missing

lookout/style/format/rules.py
```diff
@@ -19,21 +19,65 @@
...
 
 RuleStats = NamedTuple("RuleStats", (("cls", int), ("conf", float)))
 """
 `cls` is the predicted class
 `conf` is the rule confidence \\in [0, 1], "1" means super confident
 """
 
-Rule = NamedTuple("RuleType", (("attrs", Tuple[RuleAttribute, ...]), ("stats", RuleStats)))
+
+class Rule(NamedTuple("RF", (("attrs", Tuple[RuleAttribute, ...]), ("stats", RuleStats)))):
```
vmarkovtsev: Nice extension :+1:

lookout/style/format/rules.py
```diff
@@ -19,21 +19,65 @@
...
+    def humanize(self, feature_names: Sequence[str], reserved: Sequence[str]) -> str:
+        """
+        Converts Rule to human-readable format.
```
vmarkovtsev: to human-readable text description

lookout/style/format/analyzer.py
zurk: https://github.com/src-d/style-analyzer/pull/200

lookout/style/format/analyzer.py
zurk: https://github.com/src-d/style-analyzer/pull/200

lookout/style/format/features.py
zurk: it is outdated. we do not need it anymore

PR #192, opened 2018-09-30
Fix for 'missed some imaginary tokens` for regexp

zurk:
Fixes https://github.com/src-d/style-analyzer/issues/180
Also, add a new test with code in the issue.

### DIFF
--- a/lookout/style/format/features.py
+++ b/lookout/style/format/features.py
@@ -209,7 +209,6 @@ class FeatureExtractor:
         except ImportError:
             # It's normal for some languages not to have a uast_fixes module.
             self.node_fixtures = {}
-            pass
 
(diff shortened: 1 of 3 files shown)


m09:
Even if it is a temporary workaround, it should be part of a module in langs as I said in my previous comment (like `token_unwrappers`, which is also a workaround bblfsh limitations). We have to keep in mind that temporary workarounds often end up staying there for the complete life of the soft so it's better to contain them and make them maintainable as much as possible.

zurk:
ok. give me 20 mins.

zurk:
travis is green actually: https://travis-ci.org/src-d/style-analyzer/builds/435638961?utm_source=github_status&utm_medium=notification

PR #194, opened 2018-10-01
More statistics about rules & prettify 

EgorBu:
Depends on https://github.com/src-d/style-analyzer/pull/193
Before:
```
Rule 0: RuleType(attrs=(RuleAttribute(feature=1, cmp=True, threshold=1.5), RuleAttribute(feature=8, cmp=True, threshold=1.0), RuleAttribute(feature=225, cmp=True, threshold=1.0), RuleAttribute(feature=269, cmp=True, threshold=118.0)), stats=RuleStats(cls=8, conf=0.9916667, support=60.0))
Rule 1: RuleType(attrs=(RuleAttribute(feature=1, cmp=True, threshold=1.5), RuleAttribute(feature=8, cmp=True, threshold=1.0), RuleAttribute(feature=225, cmp=True, threshold=1.0), RuleAttribute(feature=269, cmp=False, threshold=118.0), RuleAttribute(feature=269, cmp=True, threshold=114.5)), stats=RuleStats(cls=7, conf=0.9992, support=625.0))
```
After:
```
Rule 0: node_start_col > 1.5 & left_1_keyword_id > 1.0 & left_3_EXPRESSION > 1.0 & left_3_STATEMENT > 118.0 gives class " with confidence 0.9916667 and support 60.0
Rule 1: node_start_col > 1.5 & left_1_keyword_id > 1.0 & left_3_EXPRESSION > 1.0 & left_3_STATEMENT <= 118.0 & left_3_STATEMENT > 114.5 gives class ' with confidence 0.9992 and support 625.0
```

Example of report in details.

<details>

```console
egor@egor-sourced:~/workspace/style-analyzer$ python3 -m lookout.style.format rule -i "/home/egor/workspace/tmp/freeCodeCamp_no_min.js/**/*" -m /tmp/home/egor/workspace/tmp/freeCodeCamp_no_min.js/style.format.analyzer.FormatAnalyzer_1.asdf
/usr/local/lib/python3.5/dist-packages/sklearn/ensemble/weight_boosting.py:29: DeprecationWarning: numpy.core.umath_tests is an internal NumPy module and should not be imported. It will be removed in a future NumPy release.
  from numpy.core.umath_tests import inner1d
[...]

PR #199, opened 2018-10-02
Pretty comment message for debugging

zurk:
Ability to get comments like this:

----
format: style mismatch:
```javascript
17|function getCompletedCertCount(user) {
18|  return
19|  [
```
Replace `<newline>` with `<space>` at column 9.
Confidence: 0.896
Triggered rule # 65:
```
    left_2_reserved in |.|
    left_1_label in |<newline>|
    left_0_reserved in |;˙{˙}|
    left_0_label in |<+space>˙<-space>˙<newline>|
    left_2_roles in |string|
    left_0_roles in |string|
```
----

This code requires more work and some refactoring. But it is hard to do in one PR at once. 
So I left todos there. It is better to start from something at least.

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -14,6 +14,8 @@ from lookout.core.api.service_analyzer_pb2 import Comment
 from lookout.core.api.service_data_pb2 import File
 from lookout.core.api.service_data_pb2_grpc import DataStub
 from lookout.core.data_requests import with_changed_uasts_and_contents, with_uasts_and_contents
+from lookout.style.format.descriptions import (
+    rule_to_comment, get_code_chunk, get_error_description)
 from lookout.style.format.diff import find_new_lines
 from lookout.style.format.feature_extractor import FeatureExtractor
 from lookout.style.format.model import FormatModel
@@ -44,6 +46,22 @@ class FormatAnalyzer(Analyzer):
         :param data: Contains "changes" - the list of changes in the pointed state.
         :return: List of comments.
         """
+        def group_line_nodes(y, y_pred, vnodes_y, rule_winners):
+            line_nodes = []
+            generate_comment = False
+            for yi, y_predi, vnode_y, winner in zip(y, y_pred, vnodes_y, rule_winners):
+                if not line_nodes or vnode_y.start.line == line_nodes[0][2].start.line:
+                    # collect all nodes on the same line
+                    line_nodes.append((yi, y_predi, vnode_y, winner))
+                    if yi != y_predi:
+                        generate_comment = True
+                    continue
+                else:
+                    if generate_comment:
+                        yield line_nodes
+                    generate_comment = yi != y_predi
+                    line_nodes = [(yi, y_predi, vnode_y, winner)]
+
         comments = []
         changes = list(data["changes"])
         base_files_by_lang = self._files_by_language(c.base for c in changes)
@@ -54,8 +72,7 @@ class FormatAnalyzer(Analyzer):
                                  len(head_files), lang, lang)
                 continue
             rules = self.model[lang]
-            for file in self._filter_files(head_files,
-                                           rules.origin_config["line_length_limit"]):
+            for file in self._filter_files(head_files, rules.origin_config["line_length_limit"]):
                 try:
                     prev_file = base_files_by_lang[lang][file.path]
                 except KeyError:
@@ -65,23 +82,51 @@ class FormatAnalyzer(Analyzer):
                 fe = FeatureExtractor(language=lang, **rules.origin_config["feature_extractor"])
                 res = fe.extract_features([file], lines)
                 if res is None:
-                    comment = Comment()
-                    comment.file = file.path
-                    comment.confidence = 100
-                    comment.line = 1
-                    comment.text = "Failed to parse this file"
+                    if self.config["report_parse_failures"]:
+                        comment = Comment()
+                        comment.file = file.path
+                        comment.confidence = 100
+                        comment.line = 1
+                        comment.text = "Failed to parse this file"
+                        comment.append(comment)
+                    self.log.warning("Failed to parse %s", file.path)
                     continue
-                X, y, vnodes_y, _ = res
-                self.log.debug("predicting values for %d samples", len(y))
-                y_pred, winners = rules.predict(X, True)
+                X, y, vnodes_y, vnodes = res
+                y_pred, rule_winners = rules.predict(X, True)
                 assert len(y) == len(y_pred)
 
-                for yi, y_predi, vnode, winner in zip(y, y_pred, vnodes_y, winners):
-                    if yi != y_predi:
-                        comment = vnode.to_comment(y_predi)
-                        comment.file = file.path
-                        comment.confidence = int(round(rules.rules[winner].stats.conf * 100))
-                        comments.append(comment)
+                code_lines = file.content.decode("utf-8", "replace").splitlines()
+                for line_nodes in group_line_nodes(y, y_pred, vnodes_y, rule_winners):
+                    code_line_number = line_nodes[0][2].start.line  # 1-based
+                    if self.config["report_triggered_rules"]:
+                        code_text = ""
+                        if self.config["report_code_lines"]:
+                            code_text = get_code_chunk(lang, code_lines, code_line_number)
+                        vnodes_comments = [
+                            "%s\n%s" % (get_error_description(vnode, y_predi),
+                                        rule_to_comment(rules.rules[winner], fe, winner))
+                            for yi, y_predi, vnode, winner in line_nodes if yi != y_predi]
+                        text = "format: style mismatch:\n%s%s\n" % (
+                            code_text, "\n\n".join(vnodes_comments))
+                    else:
+                        vnodes_comments = [
+                            get_error_description(vnode, y_predi)
+                            for yi, y_predi, vnode, winner in line_nodes if yi != y_predi]
+                        text = "format: style mismatch:\n%s\n" % ("\n\n".join(vnodes_comments))
+
+                    confidence = 0
+                    confidence_count = 0
+                    for yi, y_predi, _, winner in line_nodes:
+                        if yi != y_predi:
+                            confidence += rules.rules[winner].stats.conf
+                            confidence_count += 1
+
+                    comment = Comment()
+                    comment.line = code_line_number
+                    comment.text = text
+                    comment.file = file.path
+                    comment.confidence = int(round(confidence * 100 / confidence_count))
+                    comments.append(comment)
         return comments
 
     @classmethod
@@ -188,7 +233,9 @@ class FormatAnalyzer(Analyzer):
         :return: Full config.
         """
         defaults = {
-            "debug": False,
+            "report_code_lines": True,
+            "report_triggered_rules": True,
+            "report_parse_failures": True,
         }
         return merge_dicts(defaults, config)
 
--- a/lookout/style/format/descriptions.py
+++ b/lookout/style/format/descriptions.py
@@ -3,7 +3,7 @@ from collections import defaultdict
 from copy import copy
 from functools import singledispatch
 from math import ceil, floor
-from typing import List, Tuple
+from typing import List, Optional, Tuple, Sequence
 
 from numpy import flatnonzero, floating, ndarray
 
@@ -11,6 +11,7 @@ from lookout.style.format.feature_extractor import FeatureExtractor, FEATURES_MA
 from lookout.style.format.feature_utils import (
     CLASS_INDEX, CLASSES, CLS_DOUBLE_QUOTE, CLS_NEWLINE, CLS_NOOP, CLS_SINGLE_QUOTE, CLS_SPACE,
     CLS_SPACE_DEC, CLS_SPACE_INC, CLS_TAB, CLS_TAB_DEC, CLS_TAB_INC)
+from lookout.style.format.feature_utils import VirtualNode
 from lookout.style.format.features import BagFeature, CategoricalFeature, OrdinalFeature
 from lookout.style.format.rules import Rule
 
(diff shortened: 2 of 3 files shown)


zurk:
Also, talk to @EgorBu and he is fine with my Pretty rule formatting. He implemented his own here: https://github.com/src-d/style-analyzer/pull/194

lookout/style/format/analyzer.py
```diff
@@ -43,6 +44,43 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
         :param data: Contains "changes" - the list of changes in the pointed state.
         :return: List of comments.
         """
+        def vnode_to_comment(vnode: VirtualNode, correct_y: int) -> str:
```
vmarkovtsev: `vnode_to_error_description`

Even though we document everywhere that we return a string, everybody will still expect this func to return a Comment

lookout/style/format/analyzer.py
```diff
@@ -43,6 +44,43 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+                else:
+                    if generate_comment:
+                        yield line_nodes
+                    generate_comment = yi != y_predi
+                    line_nodes = [(yi, y_predi, vnode_y, winner)]
+
+        def get_code_chunk(lang, code_lines, line_number):
+            lines = list(range(max(0, line_number - 2), line_number + 1))
+            code_chunk = "\n".join("%d| %s" % (l, code_lines[l]) for l in lines)
+            return "```%s\n%s\n```\n" % (lang, code_chunk)
```
vmarkovtsev: I am starting to get a feeling that we need Jinja2 here. Just note for the future.

lookout/style/format/analyzer.py
```diff
@@ -64,24 +101,46 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
                 fe = FeatureExtractor(language=lang, **rules.origin_config["feature_extractor"])
                 res = fe.extract_features([file], lines)
                 if res is None:
-                    comment = Comment()
-                    comment.file = file.path
-                    comment.confidence = 100
-                    comment.line = 1
-                    comment.text = "Failed to parse this file"
+                    if debug:
```
vmarkovtsev: How about enabling the debug features explicitly, for example `report_errorred_files` instead of `debug`

lookout/style/format/analyzer.py
```diff
@@ -64,24 +101,46 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+                        vnodes_comments = [
+                            "%s\n" % (vnode_to_comment(vnode, y_predi))
+                            for yi, y_predi, vnode, winner in line_nodes if yi != y_predi]
+                        text = "format: style mismatch:\n%s\n" % ("\n".join(vnodes_comments))
+
+                    comment = Comment()
+                    comment.line = code_line_number
+                    comment.text = text
+                    comment.file = file.path
+                    comment.confidence = 100
```
vmarkovtsev: Is it possible to put the rule confidence here

lookout/style/format/rules.py
```diff
@@ -19,21 +19,74 @@
...
 `cls` is the predicted class
 `conf` is the rule confidence \\in [0, 1], "1" means super confident
 """
 
-Rule = NamedTuple("RuleType", (("attrs", Tuple[RuleAttribute, ...]), ("stats", RuleStats)))
+
+class Rule(NamedTuple("RF", (("attrs", Tuple[RuleAttribute, ...]), ("stats", RuleStats)))):
+    def humanize(self, feature_names: Sequence[str]) -> str:
+        """
+        Converts Rule to human-readable text description.
```
vmarkovtsev: New line after the description is needed

lookout/style/format/rules.py
```diff
@@ -19,21 +19,74 @@
...
+                repr[name]["str"] += "%s not in |%s|" % (name, sep.join(sorted(val["notin"])))
+
+        simple = "\n    ".join(
+            attr.humanize(feature_names[attr.feature])
+            for attr in self.attrs if not check_ohe(feature_names[attr.feature]))
+        ohe = "\n    ".join(vals["str"] for key, vals in repr.items())
+
+        return "    " + (simple + "\n    " if simple else "") + ohe
+
+    def to_comment(self, number: int, feature_names: Sequence[str]):
```
vmarkovtsev: Need to annotate the return type

lookout/style/format/analyzer.py
zurk: good idea. 
Sooo sad we cannot use fstrings for two reasons.

lookout/style/format/analyzer.py
zurk: mm, since we move out it from `VirtualNode` class I think we can shorten and name it just `error_description` WDYT?

vmarkovtsev: There must be a verb either at the beginning or at the end

vmarkovtsev:
@zurk conflicts here

vmarkovtsev:
@m09 PTAL

lookout/style/format/analyzer.py
```diff
@@ -43,6 +45,54 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
         :param data: Contains "changes" - the list of changes in the pointed state.
         :return: List of comments.
         """
+        def get_error_description(vnode: VirtualNode, correct_y: int) -> str:
```
m09: Please rename `correct_y` to `predicted_y`or `y_pred` (to match your other function) to better make the distinction between the label (`node.y`) and the prediction.

lookout/style/format/analyzer.py
```diff
@@ -43,6 +45,54 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+                        generate_comment = True
+                    continue
+                else:
+                    if generate_comment:
+                        yield line_nodes
+                    generate_comment = yi != y_predi
+                    line_nodes = [(yi, y_predi, vnode_y, winner)]
+
+        def get_code_chunk(lang, code_lines, line_number):
+            lines = list(range(max(0, line_number - 2), line_number + 1))
```
m09: should be line_number + 2 as stopping condition so that the line is in the middle.

lookout/style/format/analyzer.py
```diff
@@ -43,6 +45,54 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+        def to_comment(rule: Rule, number: int, feature_extractor: FeatureExtractor) -> str:
+            """
+            Returns the comment for the rule.
+
+            :param rule: the rule to convert to string.
+            :param number: Rule number.
+            :param feature_extractor: Corresponding feature extractor.
+            :return: String comment.
+            """
+            return "Confidence: %.3f\nTriggered rule # %d:\n```\n%s\n```" % (
```
m09: rule description already has support and confidence information

lookout/style/format/analyzer.py
```diff
@@ -64,23 +113,50 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+                        comment.line = 1
+                        comment.text = "Failed to parse this file"
+                        comment.append(comment)
+                    self.log.warning("Failed to parse %s", file.path)
                     continue
-                X, y, vnodes_y, _ = res
-                self.log.debug("predicting values for %d samples", len(y))
-                y_pred, winners = rules.predict(X, True)
+                X, y, vnodes_y, vnodes = res
+                X, _ = fe.select_features(X, y)
```
m09: it's not necessary to call select_features anymore

lookout/style/format/descriptions.py
```diff
@@ -66,7 +66,7 @@ def describe_rule(rule: Rule, feature_extractor: FeatureExtractor) -> str:
         for group, nodes in grouped.items()
         for node, feature_names in nodes.items()
         for feature_name, splits in feature_names.items()]
-    return "%s\n\t→ y = %s (%.2f confidence, %d support)" % (
+    return "\t%s\n\t→ y = %s (%.2f confidence, %d support)" % (
```
m09: I think it makes more sense to keep the first line not indented in general (and to add the tab yourself if you want it indented in your special case). Ie in rule_stats, the web viewer etc this will deteriorate the result.

lookout/style/format/analyzer.py
```diff
@@ -64,23 +113,50 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
-                        comments.append(comment)
+                code_lines = file.content.decode("utf-8", "replace").splitlines()
+                for line_nodes in group_line_nodes(y, y_pred, vnodes_y, rule_winners):
+                    code_line_number = line_nodes[0][2].start.line  # 1-based
+                    if self.config["report_triggered_rules"]:
+                        code_text = ""
+                        if self.config["report_code_lines"]:
+                            code_text = get_code_chunk(lang, code_lines, code_line_number)
+                        vnodes_comments = [
+                            "%s\n%s\n" % (get_error_description(vnode, y_predi),
```
m09: the `\n` is a joining character, it should probably be part of the join call rather than being here

lookout/style/format/analyzer.py
```diff
@@ -64,23 +113,50 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+                                          to_comment(rules.rules[winner], winner, fe))
+                            for yi, y_predi, vnode, winner in line_nodes if yi != y_predi]
+                        text = "format: style mismatch:\n%s%s\n" % (
+                            code_text, "\n".join(vnodes_comments))
+                    else:
+                        vnodes_comments = [
+                            "%s\n" % (get_error_description(vnode, y_predi))
```
m09: same here for `\n`

zurk:
ok, I see the review and apply it soon. 
@m09, @vmarkovtsev What do you think about moving all these auxiliary functions to `descriptions.py`?

m09:
All the auxililary functions that are about formatting strings would make sense in `descriptions.py`, so yes a good part of this PR could go in it :+1:

lookout/style/format/analyzer.py
zurk: no, because `line_number` is 1-based here, so it is:
 ```
line_number -= 1
list(range(max(0, line_number - 1), line_number + 2))`

m09: but there are only 4 lines selected right now right?

zurk: it is 3, because first is included and last excluded:
```python
In[1]:  l = list(range(1, 10))
In[2]:  i = 5
In[3]:  l[i-1:i+2]
Out[3]: [5, 6, 7]
```

lookout/style/format/analyzer.py
zurk: thanks!

lookout/style/format/analyzer.py
m09: huh lol, I really need a coffee. Sorry for the noise.

lookout/style/format/descriptions.py
zurk: ok, make sense.

lookout/style/format/analyzer.py
zurk: np :)

zurk:
About confidence. I use a product in pure probability sense. I assume that our predictions independent, so I take a product to get the final report confidence. I think it is an okeish way for now. We should not have more than 4-5 comments on one line and also we do not use this confidence at all.

We have individual confidence in the comment as usual. Yeap, I also like to have them. :)

m09:
But precisely the probability of the combination of edits will not make sense to the user. Confidence of the model will seem to drop randomly just because many edits are proposed on a single line. This doesn't make sense from a UX point of view even if it makes sense from a probabilistic POV. The confidence will be better represented by the mean.

zurk:
let's take the `min`?

m09:
I cannot see how this would be more informative than the average.

zurk:
ok, average then.

lookout/style/format/descriptions.py
```diff
@@ -184,3 +185,48 @@ def describe_rule_parts_ordinal(feature: OrdinalFeature, name: str,
...
+    Return the comment with regard to the correct node class.
+
+    :param vnode: Node to describe.
+    :param y_pred: The index of the correct node class.
+    :return: String comment.
+    """
+    column = vnode.start.col
+    if y_pred == CLASS_INDEX[CLS_NOOP]:
+        return "%s at column %d should be removed." % (CLASS_REPRESENTATIONS[vnode.y], column)
+    elif vnode.y == CLASS_INDEX[CLS_NOOP]:
```
vmarkovtsev: You don't need `else`-s after `return`-s.

lookout/style/format/descriptions.py
```diff
@@ -184,3 +185,48 @@ def describe_rule_parts_ordinal(feature: OrdinalFeature, name: str,
...
+        return "Replace %s with %s at column %d." % (
+            CLASS_REPRESENTATIONS[vnode.y], CLASS_REPRESENTATIONS[y_pred], column)
+
+
+def get_code_chunk(lang: str, code_lines: Sequence[str], line_number: int) -> str:
+    """
+    Return nice code snippet that can be inserted to github message.
+
+    :param lang: Code language.
+    :param code_lines: Code lines.
```
vmarkovtsev: This is a bad parameter description, Captain Obvious style :)

> Code split into lines, newline characters included/excluded (idk - please choose)

Regarding `lang`. We've got two naming styles: JavaScript and javascript. Which is used here?

lookout/style/format/descriptions.py
```diff
@@ -184,3 +185,48 @@ def describe_rule_parts_ordinal(feature: OrdinalFeature, name: str,
...
+    code_chunk = "\n".join("%d|%s" % (l, code_lines[l]) for l in lines)
+    return "```%s\n%s\n```\n" % (lang, code_chunk)
+
+
+def rule_to_comment(rule: Rule, number: int, feature_extractor: FeatureExtractor) -> str:
+    """
+    Return the comment for the rule.
+
+    :param rule: The rule to convert to string.
+    :param number: Rule number.
```
vmarkovtsev: What is "rule number"? Is it the count of the rules? Is it the index? If it's an index, then where?

vmarkovtsev:
Cowardly refusing to merge without @m09's approval.

lookout/style/format/analyzer.py
m09: ping :)

zurk: it shows in the original thread. :)
Strange, I think I fixed it... Give me 5 mins.

zurk:
Now it is addressed. Hope that is all :)

PR #210, opened 2018-10-05
Match CLASSES order when declaring individual classes in feature_utils

warenlg:
1st commit fixes a bug in the repr method of VirtualNode that causes TypeErrors when extracting the features
2nd commit is just to avoid confusion reading the classes and their corresponding indices

### DIFF
--- a/lookout/style/format/feature_utils.py
+++ b/lookout/style/format/feature_utils.py
@@ -42,7 +42,7 @@ class VirtualNode:
 
     def __repr__(self) -> str:
         return ("VirtualNode(\"%s\", y=%s, start=%s, end=%s, node=%s, path=\"%s\", "
-                "global_index=%d, labeled_index=%d)" % (
+                "global_index=%s, labeled_index=%s)" % (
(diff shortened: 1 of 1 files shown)


m09:
The repr thing won't be needed after #204 since those indices will be removed.

lookout/style/format/feature_utils.py
```diff
@@ -144,12 +144,12 @@ def to_comment(self, correct_y: int) -> Comment:
 
 
 CLS_SPACE = "<space>"
+CLS_TAB = "<tab>"
+CLS_NEWLINE = "<newline>"
```
vmarkovtsev: Why do we need to reorder these? cc @m09

m09: I think @warenlg want to reorder them so that they match the order in CLASSES (just for ease of reading/understanding).

m09:
I merged master into this branch. The bugfix isn't necessary anymore so I also changed the title to better reflect what the PR is now doing.

warenlg:
Yep thanks hugo !

PR #213, opened 2018-10-08
Add doc checks to CI (continuation of #211)

m09:


Makefile
```diff
@@ -0,0 +1,6 @@
+check:
+	flake8 --config .flake8-code .
+	pylint lookout
```
vmarkovtsev: Why do we need pylint? It is noisy as hell and much harder to maintain for a large team compared to flake8.

.flake8-doc
```diff
@@ -1,5 +1,6 @@
 [flake8]
-select=D,I
+ignore=D301
+select=D
```
vmarkovtsev: I wonder what these new options mean?

Makefile
vmarkovtsev: I would rather exclude it as in my experience it brings too much overhead for little benefits.

m09: so here I use it for only a few select errors (as can be seen in .pylintrc I exclude all, and then add just a few). It's just that pylint's docparams plugin seems to be the best option available to detect missing/faulty params in docstrings.

.flake8-doc
m09: D are all options that are about doc from the flake8-docstrings package.

D301 tells you to add r in front of your docstring if you have backslashes in it. In your case we do want to use it in a normal (not raw) string so I disabled it.

(I was just something from the WIP yesterday to detect fault params but pylint is better in that regard).

PR #217, opened 2018-10-09
Make tests great again

EgorBu:
Tests for quality_report:
* wrong path to model
* empty input
* normal run
* reuse trained rules to avoid multiple trainings

### DIFF
--- a/lookout/style/format/tests/test_quality_report.py
+++ b/lookout/style/format/tests/test_quality_report.py
@@ -1,0 +1,104 @@
+import io
+import os
+import glob
+from pathlib import Path
+import shutil
+import sys
+import tarfile
+import tempfile
+import unittest
+
+from lookout.core.tests.server import run as launch_server, find_port
+from lookout.style.format.tests.test_analyzer_integration import TestAnalyzer
+from lookout.style.format.quality_report import quality_report
+
+
+class PretrainedModelTests(unittest.TestCase):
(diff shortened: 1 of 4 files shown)


lookout/style/format/tests/test_quality_report.py
```diff
@@ -0,0 +1,155 @@
...
+        self.process = Thread(target=main)
+        try:
+            self.process.start()
+            sync.wait()
+        finally:
+            EventListener.block = block
+
+    def train(self):
+        if self.model_path is None:
+            launch_server(
```
vmarkovtsev: Can you do it in setUpClass? I see that there is an assertion but it shouldn't be much problem to remove it or move to setUp.

EgorBu: I think it's not possible without a lot of copy-paste of code
1) model is trained only one time (that's why `self.model_path is None` was introduced)
2) training of the model requires `setUp` to prepare an environment
So there is no computational overhead (model is trained once) + it requires to copy whole `setUp` to train the model in `setUpModule`. I prefer to keep it as it's.

vmarkovtsev: You already have setupclass so we should not have setupmodule at all. Besides, all the code which is executed once must go under the static calls. Waiting for the fix of this, sorry, but otherwise is blocked.

PR #222, opened 2018-10-09
Switch order for pylint and flake documentation checks.

zurk:
pylint takes a long time, so I want to ctrl+C it sometimes.
Signed-off-by: Konstantin Slavnov <kslavnov@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 check:
 	flake8 --config .flake8-code .
-	pylint lookout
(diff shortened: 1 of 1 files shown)


m09:
Hum, it's super slow for me too now while it wasn't few days ago. There should be some parameter that I dropped that changed the speed somehow, I'll try to find it. In the meanwhile we can merge this.

PR #231, opened 2018-10-11
Fix setup and Dockerfile

zurk:
Setup unable to find `README.md` because it runs from another directory.

See also: https://hub.docker.com/r/srcd/style-analyzer/builds/


zurk:
Yes, @vmarkovtsev I should fix it everywhere now :)

PR #233, opened 2018-10-11
Fix Dockerfile

zurk:
Add missing README.md file

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -24,15 +24,9 @@ echo\n\' > /browser && \
 ENV BROWSER=/browser \
     LC_ALL=en_US.UTF-8
 
-COPY lookout/core/api lookout/core/api
-COPY lookout/core/*.py lookout/core/
-COPY lookout/core/langs lookout/core/langs
-COPY lookout/style lookout/style/
-COPY lookout/__init__.py lookout/__init__.py
-COPY lookout/__main__.py lookout/__main__.py
-COPY setup.py setup.py
-COPY requirements.txt requirements.txt
-RUN pip3 install -r requirements.txt && \
+COPY . style-analyzer
+RUN cd style-analyzer && \
+    pip3 install -r requirements.txt && \
     pip3 install -e . && \
     rm -rf /usr/local/lib/python3.6/dist-packages/pyspark/
 ENTRYPOINT ["analyzer"]
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,27 @@
+current_dir = $(shell pwd)
+
+.PHONY: check
 check:
 	flake8 --config .flake8-code .
 	flake8 --config .flake8-doc .
 	pylint lookout
 
-.PHONY: check
+.PHONY: docker-build
+docker-build:
+	docker build -t srcd/style-analyzer .
+
+.PHONY: docker-test
+docker-test:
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -32,6 +32,7 @@ COPY lookout/__init__.py lookout/__init__.py
 COPY lookout/__main__.py lookout/__main__.py
 COPY setup.py setup.py
 COPY requirements.txt requirements.txt
+COPY README.md README.md
```
vmarkovtsev: We need to change the way we build this, otherwise such bugs will continue to happen.
Looks like we need to copy everything and remove those which is not relevant (e.g., the lookout sdk submodule).

zurk: ok!

zurk: we can use `.dockerignore` file for this.

Makefile
```diff
@@ -1,6 +1,31 @@
...
+docker-build: Dockerfile
+	docker build -t srcd/style-analyzer .
+
+.PHONY: docker-test
+docker-test:
+	$(MAKE) _docker-test || docker rm -f style_analyzer_bblfshd
+
+.PHONY: _docker-test
+_docker-test:
+	docker run -d --rm --name style_analyzer_bblfshd \
```
vmarkovtsev: It is either `-d` or `--rm` but not together.

Makefile
```diff
@@ -1,6 +1,31 @@
...
+		--privileged -p 9432\:9432 bblfsh/bblfshd\:v2.5.0
+	docker exec style_analyzer_bblfshd \
+		bblfshctl driver install javascript docker://bblfsh/javascript-driver\:v1.2.0
+	docker run --rm -it --network host \
```
vmarkovtsev: We should build an image with tests included rather than mount the directories inside. Much stuff is hardcoded, will be complex to maintain. So:

```make
docker-test: docker-build
    docker ps | grep bblfshd  # do not launch bblfshd each time, require it
    docker run --rm -it --network host --entrypoint python3 -w /style-analyzer srcd/style-analyser -m unittest discover
```

zurk:
@vmarkovtsev why do you like to keep `.pyc` files? are they platform independent? what happens if I add them from mac?

vmarkovtsev:
@zurk pyc files are created on Python 2.7.

zurk:
@vmarkovtsev not really, https://docs.python.org/3.6/library/py_compile.html#py_compile.compile
if I get it right, they just moved them to `__pycache__` directory.

Makefile
zurk: Actually, it is possible since https://github.com/moby/moby/pull/20848 but we do not need `--rm` here. I can remove it.

vmarkovtsev:
Yes exactly, I meant that pyc were created in the same dir in <=2.7.

Makefile
vmarkovtsev: Oh nice they finally added this feature. Indeed since we manually stop it `--rm` is not needed.

zurk:
Sorry, Vadim, I overwrite your changes. Otherwise, commits would be ugly.

zurk:
done 🎉 

PS: Do we want to automate such merges (when everything is approved and checks are passed?)

vmarkovtsev:
This is something which should always be done manually @zurk We can however automate pinging PRs which are all green.

Issue #172, opened 2018-09-27
add a test to cover asdf serialization error


zurk:
Just a reminder for myself
https://github.com/src-d/style-analyzer/pull/171

PR #234, opened 2018-10-11
Add check for analyzer model save after training

zurk:
Fixes https://github.com/src-d/style-analyzer/issues/172

### DIFF
--- a/lookout/style/format/tests/test_analyzer.py
+++ b/lookout/style/format/tests/test_analyzer.py
@@ -3,6 +3,7 @@ import logging
 import lzma
 from pathlib import Path
 import tarfile
+from tempfile import TemporaryFile
 from typing import Dict, Iterable, NamedTuple, Optional
 import unittest
 
@@ -12,6 +13,7 @@ from lookout.core.analyzer import ReferencePointer
 from lookout.core.api.service_data_pb2 import File
 from lookout.style.format.analyzer import FormatAnalyzer
 from lookout.style.format.model import FormatModel
+from lookout.style.format.tests.test_model import compare_models
 
 Change = NamedTuple("Change", [("base", File), ("head", File)])
 
@@ -81,6 +83,12 @@ class AnalyzerTests(unittest.TestCase):
         model2 = FormatAnalyzer.train(self.ptr, config, datastub)
         self.assertEqual(model1["javascript"].rules, model2["javascript"].rules)
         self.assertGreater(len(model1["javascript"]), 10)
+        # Check that model can be saved without problems and then load back
+        with TemporaryFile(prefix="analyzer_model-", suffix=".asdf") as f:
+            model2.save(f)
+            f.seek(0)
+            model3 = FormatModel().load(f)
+            compare_models(self, model2, model3)
 
     def test_analyze(self):
         common = self.base_files.keys() & self.head_files.keys()
--- a/lookout/style/format/tests/test_model.py
+++ b/lookout/style/format/tests/test_model.py
@@ -7,6 +7,26 @@ from lookout.style.format.rules import TrainableRules
 from lookout.style.format.tests.test_rules import load_abalone_data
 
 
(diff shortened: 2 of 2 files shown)


lookout/style/format/tests/test_analyzer.py
```diff
@@ -81,6 +82,9 @@ def test_train(self):
         model2 = FormatAnalyzer.train(self.ptr, config, datastub)
         self.assertEqual(model1["javascript"].rules, model2["javascript"].rules)
         self.assertGreater(len(model1["javascript"]), 10)
+        # Check that model can be saved without problems and then load back
+        with TemporaryFile(prefix="analyzer_model-") as f:
+            model2.save(f)
```
m09: we should probably load the model back to make sure everything is fine (and to sync with the comment). We can maybe share code with test_model to do so :)

vmarkovtsev: +1

lookout/style/format/tests/test_model.py
```diff
@@ -7,6 +7,26 @@
 from lookout.style.format.tests.test_rules import load_abalone_data
 
 
+def compare_models(test_case: unittest.TestCase,
+                   format_model1: FormatModel,
+                   format_model2: FormatModel):
+    test_case.assertEqual(format_model1.languages, format_model2.languages)
+    for lang in format_model1.languages:
```
EgorBu: should we add a comparison operator to `FormatModel`?

zurk: it is hard to tell because you can want different things. If you want to find out that models are from the same source, you should check metadata. if you want to check if they have the same rules, you should check rules only. And for parameters equality, you check them also. So, IMO, it is better to keep it without comparison operator until we have the real case for it.

zurk:
there is a weird error in Travis:
```
======================================================================
ERROR: test_train (lookout.style.format.tests.test_analyzer.AnalyzerTests)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/asdf/fits_embed.py", line 19, in <module>
    from astropy.io import fits
ImportError: No module named 'astropy'
During handling of the above exception, another exception occurred:
Traceback (most recent call last):
  File "/home/travis/build/src-d/style-analyzer/lookout/style/format/tests/test_analyzer.py", line 89, in test_train
    model3 = FormatModel().load(f)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/modelforge/model.py", line 106, in load
    with asdf.open(source) as model:
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/asdf/asdf.py", line 778, in open
    ignore_missing_extensions=ignore_missing_extensions)
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/asdf/asdf.py", line 671, in _open_impl
    from . import fits_embed
  File "/home/travis/virtualenv/python3.5.6/lib/python3.5/site-packages/asdf/fits_embed.py", line 22, in <module>
    raise ImportError("AsdfInFits requires astropy")
ImportError: AsdfInFits requires astropy
----------------------------------------------------------------------
```

and I do exactly the same as earlier in other tests. Need some time to investigate it.

vmarkovtsev:
@zurk This error happens when you try to load something which is not ASDF. It tries to open it as FITS and then breaks miserably because FITS is an astronomic format and `astropy` is not installed.

vmarkovtsev:
@zurk https://github.com/spacetelescope/asdf/pull/562

m09:
Maybe a `f.seek(0)` before loading would do the trick? I see that you were using the names of the tempfile before instead of the file object itself.

zurk:
oh! I think it should help.

zurk:
thanks, @m09!

m09:
@vmarkovtsev LGTM

PR #235, opened 2018-10-11
Fix for logging

EgorBu:
Handle dictionary-like & sequence-like arguments.

### DIFF
--- a/lookout/core/slogging.py
+++ b/lookout/core/slogging.py
@@ -7,7 +7,7 @@ import re
 import sys
 import threading
 import traceback
-from typing import Union
+from typing import Union, Dict, Sequence
 
 import numpy
 import xxhash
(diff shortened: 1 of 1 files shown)


lookout/core/slogging.py
```diff
@@ -60,7 +60,12 @@ def getMessage(self):
             msg = str(self.msg)
         if self.args:
             a2s = self.array2string
-            args = tuple((a2s(a) if isinstance(a, numpy.ndarray) else a) for a in self.args)
+            if isinstance(self.args, Dict):
+                args = {k: (a2s(v) if isinstance(v, numpy.ndarray) else v)
+                        for (k, v) in self.args.items()}
+            elif isinstance(self.args, Sequence):
+                args = tuple((a2s(a) if isinstance(a, numpy.ndarray) else a)
+                             for a in self.args)
```
vmarkovtsev: Otherwise if args is neither dict or sequence we should crash properly

EgorBu: Don't we expect any other type of objects?

vmarkovtsev: It is either *args or **kwargs, but who knows... There can be a crazy bug somewhere.

PR #236, opened 2018-10-11
Add docker setup for the visualizer

m09:
I only put a development docker-compose atm. It should be nice to also have a production docker-compose (probably using docker-compose.override.yml).

### DIFF
--- a/lookout/style/format/visualizer/Dockerfile.bblfsh
+++ b/lookout/style/format/visualizer/Dockerfile.bblfsh
@@ -1,0 +1,5 @@
+FROM bblfsh/bblfshd:latest
+
+RUN bblfshd \
+    && sleep 1 \
+    && bblfshctl driver install javascript bblfsh/javascript-driver:v1.2.0
--- a/lookout/style/format/visualizer/docker-compose.yml
+++ b/lookout/style/format/visualizer/docker-compose.yml
@@ -1,0 +1,41 @@
+version: "3"
+
+services:
+  bblfsh:
+    image: "srcd/style-analyzer-visualizer-bblfsh"
+    build:
+      context: .
+      dockerfile: "Dockerfile.bblfsh"
+    privileged: true
+    network_mode: "host"
+    ports:
+      - "9432:9432"
+  style-analyzer-visualizer-server:
+    image: "srcd/style-analyzer-visualizer-server"
+    depends_on:
+      - bblfsh
+    build:
+      context: "."
+      dockerfile: "Dockerfile.server"
+    network_mode: "host"
+    ports:
+      - "5001:5001"
+    volumes:
+      - "${PWD}:/visualizer"
+  style-analyzer-visualizer-client:
+    image: "srcd/style-analyzer-visualizer-client"
+    depends_on:
+      - bblfsh
+      - style-analyzer-visualizer-server
+    build:
+      context: "."
+      dockerfile: "Dockerfile.client"
+    network_mode: "host"
+    ports:
+      - "3000:3000"
+    volumes:
+      - "${PWD}/public:/visualizer/public"
+      - "${PWD}/src:/visualizer/src"
+
+volumes:
+  drivers:
--- a/lookout/style/format/visualizer/Dockerfile.client
+++ b/lookout/style/format/visualizer/Dockerfile.client
@@ -1,0 +1,11 @@
+FROM node:latest
+
+WORKDIR /visualizer
+
+COPY package.json .
+
+RUN npm install
+
+EXPOSE 3000
+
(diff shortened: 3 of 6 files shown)


lookout/style/format/visualizer/bblfsh/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM bblfsh/bblfshd:v2.9.0
+
+RUN bblfshd & sleep 1 \
+    && bblfshctl driver install javascript bblfsh/javascript-driver:v1.2.0
```
EgorBu: I think it could be a good idea to store a variable/config with drivers requirements/versions somewhere - to avoid misbehavior after some modifications in Travis for example https://github.com/src-d/style-analyzer/blob/master/.travis.yml#L67 .

lookout/style/format/visualizer/README.md
```diff
@@ -2,39 +2,16 @@
...
-    pip install -r requirements.txt
-
-## Build
-
-To build the React client, run
-
-    npm run build
-
-in the visualizer directory.
+Install [docker](https://docs.docker.com/install/) and
```
EgorBu: why not keep all options to launch visualizer? With and without docker.

m09:
Indeed you're right Egor we should document both options, I'll add that to the README :)

lookout/style/format/visualizer/bblfsh/Dockerfile
m09: alternative would be to use the image I made here (base bblfsh image + the drivers already installed) as the one used in style-analyzer in addition to the visualizer. WDYT @EgorBu @vmarkovtsev?

vmarkovtsev: For now, let's hack and slap shit together. I filed https://github.com/bblfsh/sdk/issues/317 for a proper solution.

vmarkovtsev:
@m09 We use Kubernetes instead of docker-compose, so it would be great if you refactor this to a [k8s pod](https://kubernetes.io/docs/tasks/configure-pod-container/translate-compose-kubernetes/)

m09:
@vmarkovtsev: the goal of this PR is to allow the design team to play with the visualizer with the simplest install possible. I fear that installing the k8 stack might be way worse than installing the visualizer manually.

vmarkovtsev:
Installing minicube is the same complexity as docker compose. We will use the k8s pod config not only for the product but also to run it ourselves, without building anything with npm, etc.

vmarkovtsev:
@m09 I can help you with converting the current config if you have got troubles.

vmarkovtsev:
All right, @m09 can you please split this into (everything but Dockerfile and k8s) and (Dockerfile and k8s)

m09:
@vmarkovtsev I separated the docker files from the rest.

lookout/style/format/visualizer/Dockerfile.bblfsh
```diff
@@ -0,0 +1,4 @@
+FROM bblfsh/bblfshd:latest
+
+RUN bblfshd & sleep 1 \
```
vmarkovtsev: I think it was meant to be

```
RUN bblfshd \
    && sleep 1 \
    && bblfshctl ...
```

lookout/style/format/visualizer/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+version: "3"
+
+services:
+  bblfsh:
+    image: "hugomougard/style-analyzer-visualizer-bblfsh"
```
vmarkovtsev: We should change `hugomougard` to `srcd` everywhere

lookout/style/format/visualizer/docker-compose.yml
```diff
@@ -0,0 +1,43 @@
+version: "3"
```
vmarkovtsev: This is a docker compose config while we need a pod definition. I suppose that this file is removed in the next PR.

m09:
@vmarkovtsev indeed atm it's a WIP docker config meant to be a base for further `kompose` action. As I told you in slack, I experimented with different output formats and possibilities with kompose but couldn't wrap anything up so I didn't push any output of kompose (which would have been the straight output of kompose convert).

vmarkovtsev:
@m09 ping here

m09:
@vmarkovtsev I have a setup that I feel should be working but is not. If you have any insight it's appreciated.

The hostpath mount doesn't seem to make my local file available and both the client and the server don't find the required files when running.

m09:
I'll do that but how do I then push to dockerhub?

PR #245, opened 2018-10-18
Fix bad quotes

### LINKED ISSUE #118: Fix the single quote misprediction
As shown in @EgorBu's examples, sometimes we predict pair quotes wrong - they do not match, e.g. `'test"` or `"test'`. We need to figure out why and fix this.

m09:
With this PR, we no longer predict unmatched quotes but we can be inconsistent at the document level, which is maybe not ideal still. Changing that would probably just mean hardcoding the mode of `{', "}` for each `quote string quote` triplet we detect in the ouput, which is also a bit barbaric.

Fixes #118.


lookout/style/format/feature_utils.py
```diff
@@ -11,7 +11,8 @@
 
 class VirtualNode:
     def __init__(self, value: str, start: Position, end: Position,
-                 *, node: bblfsh.Node = None, y: int = None, path: str = None) -> None:
+                 *, node: bblfsh.Node = None, y: int = None, path: str = None,
+                 labeled_index: int = None) -> None:
```
vmarkovtsev: We don't need this entire change with `labeled_index`. Build your own mapping in `postprocess` and greatly simplify the code.

m09:
@vmarkovtsev PTAL

lookout/style/format/analyzer.py
```diff
@@ -210,6 +212,23 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_request_st
         cls.log.info("trained %s", model)
         return model
 
+    @staticmethod
+    def _predict(X: numpy.ndarray, vnodes_y: Sequence[VirtualNode], vnodes: Sequence[VirtualNode],
+                 rules: Rules, language: str, return_originals: bool = False
+                 ) -> Union[Tuple[numpy.ndarray, numpy.ndarray, numpy.ndarray, numpy.ndarray],
+                            Tuple[numpy.ndarray, numpy.ndarray]]:
+        y_pred, winners = rules.predict(X, True)
```
vmarkovtsev: We do this at review time too, so everything should go under `rules.predict`

lookout/style/format/feature_extractor.py
```diff
@@ -472,6 +472,19 @@ def _inplace_write_features(features: Iterable[int], row: int, col: int, X: nump
         X[row, col:col + len(to_write)] = to_write
         return len(to_write)
 
+    def _keep_sibling(self, sibling: VirtualNode, vnode: VirtualNode, include_labeled: bool
+                      ) -> bool:
+        if not include_labeled and (sibling.y is not None
+                                    or sibling.node is None and sibling.value.isspace()):
```
vmarkovtsev: Better formatting:

```python
if not include_labeled and (
        sibling.y is not None or sibling.node is None and sibling.value.isspace()):
    #  ...
```

lookout/style/format/langs/javascript/postprocessor.py
```diff
@@ -0,0 +1,53 @@
...
+        if not i + 2 in y_indices:
+            continue
+        y_i_paired = y_indices[i + 2]
+        vnode, vnode_in_between, vnode_paired = vnodes[i], vnodes[i + 1], vnodes[i + 2]
+        if vnode.y not in quotes_classes or vnode_paired.y not in quotes_classes:
+            continue
+        if vnode_in_between.node is None:
+            continue
+        vnode_in_between_roles = {role_name(role_id) for role_id in vnode_in_between.node.roles}
+        if ("STRING" in vnode_in_between_roles
```
vmarkovtsev: Single line fits:
```python
if ("STRING" in vnode_in_between_roles and y[y_i] != y[y_i_paired]):
```

lookout/style/format/quality_report.py
```diff
@@ -32,9 +33,9 @@ def quality_report(input_pattern: str, bblfsh: str, language: str, n_files: int,
     if res is None:
         print("Failed to parse files, aborting report...")
         return
-    X, y, vnodes_y, _ = res
+    X, y, vnodes_y, vnodes = res
 
-    y_pred = rules.predict(X)
+    y_pred, _ = FormatAnalyzer._predict(X, vnodes_y, vnodes, rules, language)
```
vmarkovtsev: Hence we will not have this nasty hack here

lookout/style/format/analyzer.py
m09: rules has no knowledge of anything related to the nature of what it predicts so to me this seems like it breaks abstraction. WDYT @vmarkovtsev?

vmarkovtsev: The expectation is to take a bunch of rules and get the result of their application which you present to the users. How they are applied is completely abstracted away.

vmarkovtsev: You can look at the current code and see how twisted it becomes otherwise.

m09: BTW I can make predict public in `FormatAnalyzer` to make it clear that this is an integral part of the process that is not a hack.

vmarkovtsev: Rename the current `Rules.predict()` to `apply()` (yep, this makes sense, agree) and move `FormatAnalyzer._predict` to `Rules.predict` please.

PR #250, opened 2018-10-18
Add different styles benchmark for javascript

zurk:
this PR adds code for different styles benchmark generation. 
It creates a directory with repositories with a different style and its violation in different branches and `index.csv` file. The file looks like:
```
repo,style,from,to
jsquery,spaces_inside_round_brackets,9c161f0fbb7f1a95ff650def2a9e688d8053330f,0f4f7ae49d8f27d9310c7dcf99133e8200950515
freeCodeCamp,spaces_inside_round_brackets,6e2ce3eaa54d4039dd55a90d33d012e808498b57,ea7d587e3f37861dc6dd4211cd5177e47a9000c0
nodejs,spaces_inside_round_brackets,baeb87711ffdee63134cc3fc9b25f0699bac4483,72ea2ab0419b7480489e26d03346212f0588c404
jsquery,no_space_after_if,715bf411470620170b0c971ac17ed3ca1efb7c7f,0a1c9c30d83c15308b0d55667ea93b49e59512ec
``` 
and contains the rows with the 
1. repository name, 
2. style name, 
3. commit hash with proper style applied
4. commit hash with style violation (now it is just revert of previous commit where style was applied).

Now the code supports only JS, does not have classes and uses regexp to apply styles, but it will be easy to fix later when we actually will need all these features.


lookout/style/format/benchmarks/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+The module contains the code and data for the format style-analyzer benchmarking and benchmarks
+generation. To use it git should be installed in your system.
+"""
+# flake8: noqa: F401
+from lookout.style.format.benchmarks.generate_styles import generate_styles_entry
```
m09: We keep one to one mapping between modules and file structure everywhere so we should remove this import.

lookout/style/format/benchmarks/__main__.py
```diff
@@ -0,0 +1,65 @@
+"""CLI for benchmarking."""
```
m09: should be in `format/__main__.py` to have only one entry point to all the supported tools.

lookout/style/format/benchmarks/commit_changes.sh
```diff
@@ -0,0 +1,13 @@
...
+  echo "no changes"
+  exit 1
+fi
+git checkout -b $2 --quiet
+git commit -a -m $2 --quiet
+git rev-parse HEAD
+git revert HEAD --no-edit 1>&2
+git rev-parse HEAD
+git checkout master --quiet
+exit 0
```
m09: misses newline and it might be better to just remove `exit 0` altogether since it overrides the return value of something that should not fail.

lookout/style/format/benchmarks/generate_styles.py
```diff
@@ -0,0 +1,114 @@
...
+js_styles_regex = {
+    "equal_no_space_style": (" = ", "="),
+    "operators_no_space_style": (" (%s) " % js_operators_regex, r"\1"),
+    "no_newline_before_function": ("(\n+)function ", "\nfunction "),
+    "two_newlines_before_function": ("(\n+)function ", "\n\nfunction "),
+    "no_space_after_if": (r"if\s*\(", r"if("),
+    "no_spaces_inside_brackets": (r"\{ +([^{}]+) +\}", r"{\1}"),
+    "no_spaces_inside_round_brackets": (r"\( +([^()]+) +\)", r"(\1)"),
+    "spaces_inside_round_brackets": (r"\(([^()]+)\)", r"( \1 )"),
+}
```
m09: This may be enough for very basic changes (even though those listed here will already be super weird in the sense that they don't distinguish between function `{}`-s and objects `{}`-s, same for parens and brackets, etc) but if we want to do anything even remotely more complicated we should use ASTs to transform formatting, not regexes.

zurk:
there is also one solution I want to discuss. I decided not to add a dependency from some python-git library to work with the git repo. On the other side, I think it is first but not the last case when we need to work with git from python. We can keep it as is now but in case we have this situation one more time let's add some python-git library and reimplement `commit_changes` function. 

WDYT?

lookout/style/format/benchmarks/generate_styles.py
```diff
@@ -0,0 +1,114 @@
...
+                    from_commit, to_commit = commit_changes(repopath, style_name)
+                except _NoChangesException:
+                    log.info("No changes for the repo %s for style %s", repopath, style_name)
+                    continue
+                index_file.write("%s,%s,%s,%s\n" % (
+                    repopath.name, style_name, from_commit, to_commit))
+    return 0
+
+
+if __name__ == "__main__":
```
m09: should be removed

lookout/style/format/tests/test_behchmarks.py
```diff
@@ -0,0 +1,31 @@
+import os
```
m09: name of the file should be test_benchmarks.py

m09:
> there is also one solution I want to discuss. I decided not to add a dependency from some python-git library to work with the git repo. On the other side, I think it is first but not the last case when we need to work with git from python. We can keep it as is now but in case we have this situation one more time let's add some python-git library and reimplement `commit_changes` function.
> 
> WDYT?

I think we should use a python-only git library if one exists and works well (I didn't look into that so I hope one does exist!). Since the script is still small it should be easy to port but if not it could wait for after the 25th (in this case we should open an issue).

zurk:
@m09 I know about this one: https://github.com/dulwich/dulwich looks good, but I did not try it.

m09:
> Supported Python versions are 2.7, 3.4 and 3.5.

:disappointed:

lookout/style/format/benchmarks/generate_styles.py
zurk: Yes, but it is to make it faster now, I also do not like it. I think to have two options here: 
1. regexp for simple cases. 
2. Any patch function you want otherwise. 
If you want I can make a TODO to remember about it :)

lookout/style/format/benchmarks/commit_changes.sh
zurk: sorry, I do not get which return values it can override?

m09: in a script the last command's return value is returned (here the one of `git checkout master --quiet`). If this fails it's good info so just removing the exit is probably a net benefit.

zurk:
I think it is outdated: check Travis: https://github.com/dulwich/dulwich/blob/master/.travis.yml

m09:
Good point. And I think I recall us using it in `ml` also, so using it again would settle for a common lib, which is good (I recall it for bad reasons though, I think it broke the build or something similar around the point I joined).

lookout/style/format/benchmarks/commit_changes.sh
zurk: ok, got it

vmarkovtsev:
Dulwich was successfully used by @r0mainK in Modelforge already, so it is included into our dependencies. All the popular git functions should work.

lookout/style/format/__main__.py
```diff
@@ -90,6 +91,23 @@ def add_parser(name, help):
                                     "modified by adding artificial style mistakes.")
     add_bblfsh_arg(robust_parser)
     add_model_args(robust_parser)
+
+    # Generate different styles dataset
+    gen_styles_parser = add_parser("gen-styles-dataset",
```
vmarkovtsev: We need to find smth more precise here to not confuse with the evaluation dataset.

"gen-smoke-dataset"

In the theory of testing "smoke tests" mean super important integration tests which indicate the basic system functionality.

lookout/style/format/benchmarks/generate_styles.py
```diff
@@ -0,0 +1,113 @@
...
+    "=", "+=", "-=", "/=", "%=",
+    "==", "===", "!=", "!==", ">", "<", ">=", "<=",
+    "&", "|", "~", "^", ">>", "<<", ">>>",
+    "&&", "||", "!",
+    "?", ":",
+]
+
+js_operators_regex = "|".join(re.escape(op) for op in sorted(js_operators, reverse=True))
+
+js_styles_regex = {
```
vmarkovtsev: regexes

lookout/style/format/benchmarks/generate_styles.py
```diff
@@ -0,0 +1,113 @@
...
+
+def commit_changes(repopath: Path, style_name: str) -> Tuple[str, str]:
+    """
+    Call bash script which commit all changes to `style_name` branch and checkout master back.
+
+    :param repopath: Path to the repository for which style were applied.
+    :param style_name: Applied style name.
+    :return: Two commit hashes: where style was applied and where style was disrupt.
+    """
+    scriptpath = Path(__file__).parent / "commit_changes.sh"
```
vmarkovtsev: Yeah, this is 💯 ugly without Dulwich.

lookout/style/format/benchmarks/generate_styles.py
```diff
@@ -0,0 +1,113 @@
...
+    :param force: Override output directory if exists.
+    :param log_level: logging level.
+
+    :return: Success status. 0 if Succeeded, 1 otherwise.
+    """
+    slogging.setup(log_level, False)
+    log = logging.getLogger("styles-gen")
+    if not input.endswith(".tar.gz"):
+        raise ValueError("Input file should be .tar.gz archive.")
+    input = Path(input)
```
vmarkovtsev: `input` is a builtin function in Python, let's rename "input" and "output"

lookout/style/format/benchmarks/generate_styles.py
```diff
@@ -0,0 +1,113 @@
...
+            pattern = re.compile(pattern)
+            for repopath in repopaths:
+                for filepath in repopath.glob("**/*.js"):
+                    code = filepath.read_text()
+                    new_code = re.sub(pattern, repl, code)
+                    filepath.write_text(new_code)
+                try:
+                    from_commit, to_commit = commit_changes(repopath, style_name)
+                except _NoChangesException:
+                    log.info("No changes for the repo %s for style %s", repopath, style_name)
```
vmarkovtsev: in the repo

vmarkovtsev: for style check? (not sure, suggesting)

vmarkovtsev:
+ better to use tar.xz instead of tar.gz in the tests

zurk:
> better to use tar.xz instead of tar.gz in the tests
ok, but why?

vmarkovtsev:
2x size boost in the best case, and our repo is already drowning in tarballs.

zurk:
An issue with bash script remains. I hope I fix it in the evening with a help of @vmarkovtsev

vmarkovtsev:
@zurk This should help you:

```python
import os
from dulwich.repo import Repo
from dulwich.index import build_index_from_tree, blob_from_path_and_stat

# git clone https://github.com/src-d/hercules /tmp/hercules
repo = Repo("/tmp/hercules")
base = repo.head()

# sample changes
with open(os.path.join(repo.path, "labours.py"), "w") as fout:
    fout.write("hello, world")

# commit everything
for tree_path, entry in repo.open_index().iteritems():
    full_path = os.path.join(repo.path.encode(), tree_path)
    blob = blob_from_path_and_stat(full_path, os.lstat(full_path))
    if blob.id != entry.sha:
        repo.stage(tree_path)
repo.do_commit(b"My commit message", b"Vadim Markovtsev <vadim@sourced.tech>")

# rollback
tree = repo[base].tree
build_index_from_tree(repo.path, repo.index_path(), repo.object_store, tree)
repo[b"HEAD"] = base
```

See, dulwich is super simple and intuitive!

vmarkovtsev:
> I was able to do it with dulwich, thanks @vmarkovtsev, but wow, it is like x100 times slower.

As everything implemented in Python haha

lookout/style/format/benchmarks/generate_smoke.py
```diff
@@ -0,0 +1,115 @@
...
+        outputpath.mkdir()
+    except FileExistsError:
+        log.error("Directory %s exists. If you want to override it run with --force flag.",
+                  outputpath)
+        return 1
+    tarfile.open(str(inputpath)).extractall(str(outputpath))
+    repopaths = [x for x in outputpath.iterdir() if x.is_dir()]
+    log.info("Repos found: %s", ", ".join(r.stem for r in repopaths))
+    with open(str(outputpath / index_filename), "w") as index_file:
+        index_file.write("repo,style,from,to\n")
```
vmarkovtsev: this is why we have [`csv.writer`](https://docs.python.org/3/library/csv.html#csv.writer) in the stdlib

lookout/style/format/benchmarks/generate_smoke.py
```diff
@@ -0,0 +1,115 @@
...
+    outputpath = Path(outputpath)
+    if force and outputpath.exists():
+        shutil.rmtree(str(outputpath))
+    try:
+        outputpath.mkdir()
+    except FileExistsError:
+        log.error("Directory %s exists. If you want to override it run with --force flag.",
+                  outputpath)
+        return 1
+    tarfile.open(str(inputpath)).extractall(str(outputpath))
```
vmarkovtsev: `with tarfile.open`

zurk: why? the result of my function in a directory with index file and repositories as subdirectories.

zurk: ok, I think I got it 
```python
    with tarfile.open(str(inputpath)) as tar:
        tar.extractall(str(outputpath))
```

vmarkovtsev: yes

vmarkovtsev:
@m09 PTAL

vmarkovtsev:
@zurk https://travis-ci.org/src-d/style-analyzer/jobs/445015919#L583

zurk:
yup, additional dependencies for flake8 are installed now. :+1:

vmarkovtsev:
@zurk 
```
AssertionError: 1 is not false : You forgot to add to this test {'gen-smoke-dataset'} subcommand(s) check
```

zurk:
Ok, I finally tired to worry Travis and add `pre-commit` hook.

zurk:
this issue: https://github.com/bblfsh/client-python/issues/131
fall back to old bblfsh python-client version

PR #255, opened 2018-10-19
Re-enable import order checks

m09:
When we merged the different CI PRs, the one moving all requirements to `requirements-lint.txt` must have been rebased after the one putting import order in place and thus import order is skipped since then.


m09:
@vmarkovtsev PTAL :)

PR #260, opened 2018-10-23
Move server.py from tests to core

zurk:
I need this code to get evaluation on smoke dataset


zurk:
I will create an issue about coverage after the merge. 
@vmarkovtsev please review.

PR #262, opened 2018-10-23
Minor server.py improvements

zurk:
the same as https://github.com/src-d/style-analyzer/pull/260 but I leave the file in tests because it is not production-ready code.

### DIFF
--- a/lookout/core/tests/server.py
+++ b/lookout/core/tests/server.py
@@ -1,4 +1,6 @@
+"""Utils to work with lookout-sdk binary."""
 import io
+import logging
 import os
 import pathlib
 import random
@@ -7,6 +9,7 @@ import socket
 import subprocess
 import sys
 import tarfile
+from urllib.error import HTTPError
 from urllib.request import urlopen
 
 from lookout.core.api.version import __version__ as binver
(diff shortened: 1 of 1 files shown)


zurk:
And issue for next refactoring: https://github.com/src-d/style-analyzer/issues/263

lookout/core/tests/server.py
```diff
@@ -7,37 +9,75 @@
...
+                     "%s/lookout-sdk_%s_%s_amd64.tar.gz" % (binver, binver, sys.platform)
+                     ) as response:
             copyfileobj(response, buffer)
         buffer.seek(0)
         with tarfile.open(fileobj=buffer, mode="r:gz") as tar:
             with file.open("wb") as fout:
                 copyfileobj(tar.extractfile("lookout-sdk_linux_amd64/lookout-sdk"), fout)
         os.chmod(str(file), 0o775)
-    except Exception as e:
+        return file
```
vmarkovtsev: `file` is already present in the global namespace. Let's not prematurely overcomplicate.

lookout/core/tests/server.py
```diff
@@ -7,37 +9,75 @@
...
+    except HTTPError as e:
+        if e.code == 404:
+            log.error("Release %s for %s platform is missing." % (binver, sys.platform))
+        raise e
+    except Exception:
         if file.exists():
             os.remove(str(file))
-        raise e from None
+        raise
```
vmarkovtsev: `from None` is required! Otherwise you will spam the error trace with useless details

lookout/core/tests/server.py
```diff
@@ -7,37 +9,75 @@
...
+
+
+def run(cmd: str, fr: str, to: str, port: int, git_dir: str, config_json: str=None) -> None:
```
vmarkovtsev: We should preserve the default value of `git_dir` - as the binary itself has

lookout/core/tests/server.py
zurk: I thought it is useful...

lookout/core/tests/server.py
```diff
@@ -15,29 +18,65 @@
 file = pathlib.Path(__file__).parent / "server"
 
 
-def fetch():
+def fetch() -> pathlib.Path:
```
vmarkovtsev: Forgot to remove the type hint

lookout/core/tests/server.py
```diff
@@ -15,29 +18,65 @@
...
+    """
+    Fetch corresponding lookout-sdk executable.
+
+    :return: Path to lookout-sdk executable
```
vmarkovtsev: Here too

lookout/core/tests/server.py
```diff
@@ -15,29 +18,65 @@
...
             copyfileobj(response, buffer)
         buffer.seek(0)
         with tarfile.open(fileobj=buffer, mode="r:gz") as tar:
             with file.open("wb") as fout:
                 copyfileobj(tar.extractfile("lookout-sdk_linux_amd64/lookout-sdk"), fout)
         os.chmod(str(file), 0o775)
+    except HTTPError as e:
+        if e.code == 404:
+            log.error("Release %s for %s platform is missing." % (binver, sys.platform))
+        raise e
```
vmarkovtsev: from None

zurk:
code coverage only :thinking: 
Should be fine to merge.

PR #265, opened 2018-10-24
Add delete line surroundings to analyze

zurk:
Imagine the situation:
We have this code in base commit:
```python
class A:
    pass


class B:
    pass
```
And bad-style head commit suggests us delete one newline:
```python
class A:
    pass

class B:
    pass
```
We see only one deletion in the diff and ignore it now (that how `find_new_lines` works). So style analyzer gives No feedback.
I suggest to include one line before deletion and one line after. 

Also I rename `find_new_lines` to `find_modified_lines` for better understanding what the function do.

### DIFF
--- a/lookout/core/lib.py
+++ b/lookout/core/lib.py
@@ -27,6 +27,27 @@ def find_new_lines(before: File, after: File) -> List[int]:
     return result
 
 
+def find_deleted_lines(before: File, after: File) -> List[int]:
+    """
+    Return line numbers next to deleted lines in the new file.
+
+    :param before: the previous contents of the file.
+    :param after: the new contents of the file.
+    :return: list of line numbers next to deleted lines.
+    """
+    before_lines = before.content.decode("utf-8", "replace").splitlines()
+    after_lines = after.content.decode("utf-8", "replace").splitlines()
+    matcher = SequenceMatcher(a=before_lines, b=after_lines)
+    result = []
+    for action, _, _, j1, _ in matcher.get_opcodes():
+        if action == "delete":
+            if j1 != 0:
+                result.append(j1)
+            if j1 != len(after_lines):
+                result.append(j1 + 1)
+    return result
+
+
 def extract_changed_nodes(root: Node, lines: Sequence[int]) -> List[Node]:
     """
     Collect the list of UAST nodes which lie on the changed lines.
--- a/lookout/core/tests/test_lib.py
+++ b/lookout/core/tests/test_lib.py
@@ -3,11 +3,51 @@ import unittest
 from bblfsh import Node, Position
 
 from lookout.core.api.service_data_pb2 import File
-from lookout.core.lib import extract_changed_nodes, files_by_language, filter_files, find_new_lines
+from lookout.core.lib import (
+    extract_changed_nodes, files_by_language, filter_files, find_deleted_lines, find_new_lines)
 
 
 class LibTests(unittest.TestCase):
-    def test_diff(self):
+    def test_find_deleted_lines(self):
+        text_base = """
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+        Maecenas volutpat dui id ipsum cursus, sit amet accumsan nisl ornare.
+        Vivamus euismod lorem viverra semper dictum.
+        Nam consectetur enim eget elementum mattis.
+        Ut condimentum metus vehicula tellus tempus, vel ultricies lectus dapibus.
+        Etiam vitae nisi at ante pretium lacinia et eu massa."""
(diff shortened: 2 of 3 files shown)


m09:
We might need something a bit more complicated here. To me the best option would be to gather prediction with and without deletion and only use the comments if the predictions differ.

m09:
to reduce false positives: if we would have given the same notes without the deletion then the edit didn't change the style. We don't give predictions on the rest of the code so we should not here also.

m09:
but this is not important before release so if it takes you time to implement please create an issue for post-release time

vmarkovtsev:
@zurk What's your decision?

zurk:
Discussed in DM since I did not get the suggestion.
So Hugo's words with explanation what he suggests:

> my point is to do the following: when someone deletes 2, predict on
> 
> ```
>1
>2
>3
>```
> 
> and on
> 
>```
> 1
> 3
>```
> and compare predictions for lines 1 and 3
> if they are the same then we don't output them
> else we do

And It makes sense. We can avoid additional false-positives. But it is also required to parse the file from a base commit.

I have 3rd suggestion to include only formatting tokens close to the line. For example, if we have:
```javascript
// comment

function setPositiveNumber( elem, value, subtract ) {
    // blah
}
```
and we change it to (no new line)
```
// coment⏎
function setPositiveNumber( elem, value, subtract ) {
    // blah
}
```
We should check `⏎` only. 

Anyway, let's me create a PR about that, It will take some time to implement, I want to finish my smoke dataset evaluation ASAP.

So, can we merge this fix now @vmarkovtsev , @m09 or you prefer to wait until the issue will be solved?

m09:
I'm ok to merge as is but as I said in one of my previous comments we should also open an issue to be sure not to forget that this can be improved (and this is less important than debugging features and the rest of quality first so it should be done after release)

zurk:
done

lookout/core/lib.py
```diff
@@ -9,21 +9,29 @@
...
     :param before: the previous contents of the file.
     :param after: the new contents of the file.
-    :return: list of line numbers new to `after`.
+    :return: list of line numbers modified to `after`.
     """
-    matcher = SequenceMatcher(a=before.content.decode("utf-8", "replace").splitlines(),
-                              b=after.content.decode("utf-8", "replace").splitlines())
+    before_lines = before.content.decode("utf-8", "replace").splitlines()
+    after_lines = after.content.decode("utf-8", "replace").splitlines()
+    matcher = SequenceMatcher(a=before_lines, b=after_lines)
```
vmarkovtsev: before_lines/after_lines change is not needed and does not improve anything

lookout/core/lib.py
```diff
@@ -9,21 +9,29 @@
 from lookout.core.api.service_data_pb2 import File
 
 
-def find_new_lines(before: File, after: File) -> List[int]:
+def find_modified_lines(before: File, after: File) -> List[int]:
```
vmarkovtsev: So there is a problem with terminology here.

`find_new_lines` returns the lines which are either added or changed. IT should forever stay so.

What we should really do is to add `find_deleted_lines` which returns the deleted line indices **in the new file coordinates**. This is something not every analyzer needs, so our code will
```
sorted(chain.from_iterable(find_new_lines(), find_deleted_lines()))
```
which is equvalent to the coded `find_modified_lines`

lookout/core/lib.py
zurk: I use `after_lines` here: https://github.com/src-d/style-analyzer/pull/265/files/e0e436f5e6906d6f0fa82f2abae7ad61d7584fa5#diff-a7589e8177477e8366513b2d10ffeb6bR31

lookout/core/lib.py
zurk: ok

zurk: one thing, it requires to run `SequenceMatcher` two times and it is kind of heavy operation. Are we fine with that?

vmarkovtsev: It's not heavy compared to getting UASTs and especially evaluating our rules.

lookout/core/tests/test_lib.py
```diff
@@ -3,11 +3,12 @@
 from bblfsh import Node, Position
 
 from lookout.core.api.service_data_pb2 import File
-from lookout.core.lib import extract_changed_nodes, files_by_language, filter_files, find_new_lines
+from lookout.core.lib import (
+    extract_changed_nodes, files_by_language, filter_files, find_modified_lines)
```
vmarkovtsev: This code cannot work.

zurk: yup, solving

Issue #266, opened 2018-10-24
Split core and format python modules to two python packages.


zurk:
Now there is one: https://pypi.org/project/lookout-style

We should have something like `lookout-core` and `lookout-style`

m09:
Done.

PR #267, opened 2018-10-24
Move __version__ to style submodule

zurk:
As was discussed this `__version__` related to style-analyzer part and for the lookout-core part will be another package and module( see issue https://github.com/src-d/style-analyzer/issues/266)

### DIFF
--- a/lookout/core/__init__.py
+++ b/lookout/core/__init__.py
@@ -1,0 +1,4 @@
+# This is version for further lookout-core python package.
+# It should be used after core+format splitting and this comment should be deleted.
+# If you want to bump version now go to lookout.style.__init__.__version__
+__version__ = 0, 0, 1
--- a/lookout/style/__init__.py
+++ b/lookout/style/__init__.py
@@ -1,2 +1,3 @@
 """Supported machine learning based analyzers for lookout."""
+__version__ = 0, 0, 3
 __meta__ = True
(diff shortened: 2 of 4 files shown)


zurk:
but it is in the wrong place even now. this `__init__` can be overwritten by `lookout.sdk` (https://github.com/src-d/lookout-sdk) package.

m09:
Indeed. Then I'd rather have a `__version__.py` file in the `lookout` folder in the meantime. It's just weird to have a file affecting two packages but placed in only one of them.

lookout/core/__init__.py
```diff
@@ -0,0 +1,4 @@
+# This is version for further lookout-core python package.
+# It should be used after core+format splitting and this comment should be deleted.
+# If you want to bump version now go to lookout.style.__init__.__version__
+__version__ = (0, 0, 1)
```
vmarkovtsev: Should not be braces

lookout/style/__init__.py
```diff
@@ -1,2 +1,3 @@
 """Supported machine learning based analyzers for lookout."""
+__version__ = 0, 0, 3
```
vmarkovtsev: Can you also please remove `__meta__` because damn it is only needed in the tests as far as i remember

vmarkovtsev: Shit we need to document that

PR #284, opened 2018-10-26
Add release to Helm and Docker hub using src-d/ci

meyskens:
This will replace the docker hub builds by a Travis release that will also update the deployment in lookout staging.

This still needs the Docker credentials in Travis CI (which I don't have yet) cc @rporres

### DIFF
--- a/.helm_staging.yml
+++ b/.helm_staging.yml
@@ -1,0 +1,12 @@
+app:
+  dataServiceUrl: ipv4://lookout:10301
+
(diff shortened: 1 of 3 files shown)


rporres:
The problem is that it doesn't work completely (see https://github.com/src-d/lookout-gometalint-analyzer/pull/13 comments), reverting it. We will send a new PR when things are completely ironed.

Apologies for the noise.

PR #293, opened 2018-10-31
Improve rule descriptions

### LINKED ISSUE #282: Use signed numbers instead of left.# and right.# in the Visualizer
I suggest to print the rule parts as

```
-1.internal_type
+2.offset
```

instead of

```
left.0.internal_type
right.1.offset
```

I think it will make the rules clearer: the property names will be aligned at 3 chars, also no need to parse "left" and "right".

"parents" are not touched by this.

vmarkovtsev:
Fixes #282


vmarkovtsev:
@m09 All good?

PR #307, opened 2018-11-07
Remove dco.yml

zurk:
As I get it we decide that source{d} members should always sign-off PRs.


zurk:
I stopped Travis checks. The code is not affected.

PR #314, opened 2018-11-12
Update deps and adapt code to prevent the new type error

m09:
I also re-ran prettier, which is why the modif list is long even though the semantic changes are very small.


creachadair:
It's up to you, but I suggest you consider splitting this so that the non-semantic changes are a separate PR. It's a little extra effort up front, but simplifies reverts and makes life easier for folks searching the logs for effectful updates.

m09:
@creachadair I completely agree in general. However since this is viz code and another PR has rebased this one I'll this one slip through.

PR #316, opened 2018-11-13
Fix for missing multiline VirtualNodes

zurk:
Example:
```python
vnode.start.line = 10
vnode.end.line = 13

file_lines = [11]
```
Checking `vnode.start.line in file_lines` is not enough in this case, so we miss this VirtualNode before.

### DIFF
--- a/lookout/style/format/feature_utils.py
+++ b/lookout/style/format/feature_utils.py
@@ -1,5 +1,5 @@
 """Facilities to create and use features."""
-from typing import Callable, Iterable, Mapping, NamedTuple, Tuple, Union
+from typing import Callable, Iterable, Mapping, NamedTuple, Optional, Set, Tuple, Union
 
 import bblfsh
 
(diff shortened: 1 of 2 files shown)


m09:
Good catch!

zurk:
This is why I like my smoke dataset. I am deeply covered with our bugs :)
I think there is more

zurk:
I found out that we use the same condition twice, so move it to separate function inside VirtualNode.

lookout/style/format/feature_utils.py
```diff
@@ -61,6 +61,18 @@ def __eq__(self, other: "VirtualNode") -> bool:
...
+    def is_labeled_on_lines(self, lines: Optional[Set[int]]) -> bool:
+        """
+        Return true for labeled VirtualNode instance that located on specified lines.
+
+        :param lines: list of lines or None. None is considered as all possible lines.
+        :return: Condition value
+        """
+        if self.y is None:
+            return False
+        return (len(set(range(self.start.line, self.end.line + 1)) & lines) != 0
```
m09: return lines is None or lines.intersection(range(self.start.line, self.end.line + 1))

zurk: ok,
```
return lines is None or bool(set(range(self.start.line, self.end.line + 1)) & lines)
```
just to always have bool value.

m09: Not needed.

m09: And please use lines.intersection since it doesn't create a useless set

zurk: >And please use lines.intersection since it doesn't create a useless set

oh, sorry. I missed that. 

> Not needed.

but I have a typo hint `-> bool`
do you want my to change it to `-> Union[bool, Set]`?
I'd prefer to keep `bool(...)`

zurk:
```python
In[8]: False or set()
Out[8]: set()
```

m09: as you want, this will be rebased into one inlined call where the bool won't be present anyway.

PR #318, opened 2018-11-16
Add react repo to smoke dataset

zurk:
There is a lot of js lines in react. Helps to see performance for big repos.
This file becomes bigger and I also want to add more repos later. 
So... maybe it is worth to move it somewhere else and fetch when we need it?

Not so good to store data in git.


zurk:
The code is not affected. I stopped Travis.

vmarkovtsev:
2MB is fine for now

Issue #334, opened 2018-11-21
Use github suggestion feature


zurk:
Add to lookout comments 
```
` ` ` suggestion
<code line here>
` ` ` 
```

PR #335, opened 2018-11-21
Add code line suggestion feature

zurk:
Also, Rework FormatAnalyzer.analyze() function.
Closes https://github.com/src-d/style-analyzer/issues/334

Full debug comment now looks like:

-------
format: style mismatch:
```javascript
388|            // Multiple arguments are aggregated like Promise.all array elements
389|            while (i--) {
390|                    adoptValue( resolveValues[ i ], updateFunc(i), master.reject );
```

```suggestion
            while ( i-- ) {
```

Insert ␣ at column 15.
Triggered rule # 1:
```
        <RULE # 1 HERE>
        Confidence: 1.000. Support: 100.
```
Insert ␣ at column 18.
Triggered rule # 22:
```  
        <RULE # 2 HERE>
        Confidence: 1.000. Support: 100.
```

-------

And without debug info: 

-------
format: style mismatch:
Insert ␣ at column 15.
Insert ␣ at column 18.
```suggestion
            while ( i-- ) {
```
-------

^ this last block is line suggestion block

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -1,4 +1,5 @@
 """Analyzer that detects bad formatting by learning on the existing code in the repository."""
+import functools
 from itertools import chain
 import logging
 import os
@@ -11,13 +12,10 @@ from jinja2 import Template
 from lookout.core import slogging
 from lookout.core.analyzer import Analyzer, ReferencePointer
 from lookout.core.api.service_analyzer_pb2 import Comment
+from lookout.core.api.service_data_pb2 import File
 from lookout.core.data_requests import DataService, \
     with_changed_uasts_and_contents, with_uasts_and_contents
 from lookout.core.lib import files_by_language, filter_files, find_deleted_lines, find_new_lines
-import numpy
-from skopt import BayesSearchCV
-from skopt.space import Categorical, Integer
-
 from lookout.style.format.classes import CLASS_INDEX, CLS_NEWLINE
 from lookout.style.format.code_generator import CodeGenerator
 from lookout.style.format.descriptions import describe_rule, get_change_description
@@ -27,6 +25,9 @@ from lookout.style.format.postprocess import filter_uast_breaking_preds
 from lookout.style.format.rules import Rules, TrainableRules
 from lookout.style.format.utils import generate_comment, merge_dicts
 from lookout.style.format.virtual_node import VirtualNode
+import numpy
+from skopt import BayesSearchCV
+from skopt.space import Categorical, Integer
 
 
 class FormatAnalyzer(Analyzer):
@@ -94,7 +95,7 @@ class FormatAnalyzer(Analyzer):
         """
         super().__init__(model, url, config)
         self.config = self._load_analyze_config(self.config)
-        with open(self.config["comment_template"]) as f:
+        with open(self.config["comment_template"], encoding="utf-8") as f:
             self.comment_template = Template(f.read(), trim_blocks=True, lstrip_blocks=True)
 
     @with_changed_uasts_and_contents
@@ -142,7 +143,7 @@ class FormatAnalyzer(Analyzer):
                     log.warning("Failed to parse %s", file.path)
                     continue
                 comments.extend(self._generate_file_comments(
-                    lang, file, fe, feature_extractor_output, rules, data_service.get_bblfsh()))
+                    lang, file, fe, feature_extractor_output, data_service.get_bblfsh()))
         return comments
 
     @classmethod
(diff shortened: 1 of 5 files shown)


lookout/style/format/analyzer.py
```diff
@@ -38,6 +39,9 @@ class FormatAnalyzer(Analyzer):
         "report_code_lines": True,
         "report_triggered_rules": True,
         "report_parse_failures": True,
+        "report_github_suggestion": True,
```
vmarkovtsev: Does not need a switch, we always do this.

lookout/style/format/analyzer.py
```diff
@@ -38,6 +39,9 @@ class FormatAnalyzer(Analyzer):
...
+        "report_error_description": True,
```
vmarkovtsev: Same here.

lookout/style/format/analyzer.py
```diff
@@ -38,6 +39,9 @@ class FormatAnalyzer(Analyzer):
...
+        "uast_break_check": True,
```
vmarkovtsev: We need a Jinja2 template for our comments. There should be the default one and the ability to change it in config.

zurk:
addressed

lookout/style/format/analyzer.py
```diff
@@ -38,6 +41,9 @@ class FormatAnalyzer(Analyzer):
         "report_code_lines": True,
         "report_triggered_rules": True,
         "report_parse_failures": True,
+        "uast_break_check": True,
+        "report_template": os.path.join(os.path.dirname(__file__),
```
vmarkovtsev: Let's name it "comment_default.jinja2" and place it in "templates" directory

lookout/style/format/analyzer.py
```diff
@@ -89,6 +95,8 @@ def __init__(self, model: FormatModel, url: str, config: Mapping[str, Any]) -> N
         super().__init__(model, url, config)
         self.config = self._load_analyze_config(self.config)
         self.client = BblfshClient(self.config["bblfsh_address"])
+        with open(self.config["report_template"]) as f:
```
vmarkovtsev: report_template -> comment_template

lookout/style/format/analyzer.py
```diff
@@ -289,6 +238,100 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_request_st
...
+        confidence = 0
+        confidence_count = 0
+        for yi, y_predi, winner in zip(line_ys, line_ys_pred, line_winners):
+            if yi != y_predi:
+                confidence += rules.rules[winner].stats.conf
+                confidence_count += 1
+        return int(round(confidence * 100 / confidence_count))
+
+    @staticmethod
+    def _group_line_nodes(y, y_pred, vnodes_y, new_vnodes, rule_winners):
```
vmarkovtsev: I see that we are doing many complex actions here... Can you please add a docstring which describes the hell inside?

lookout/style/format/analyzer.py
```diff
@@ -289,6 +238,100 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_request_st
...
+                    continue
+                descriptions.append(
+                    (line_winner, get_error_description(line_vnode_y, line_y_predi, fe)))
+            text = self.report_template.render(
+                config=self.config,
+                lang=lang,
+                line_number=line_number,
+                code_lines=code_lines,
+                new_code_line=new_code_line,
+                descriptions=descriptions,
```
vmarkovtsev: We need a comment for each data item we feed to the template, e.g.
```python
text = self.report_template.render(
    config=self.config,  # configuration of the analyzer
    lang=lang,  # programming language of the code
)
```

"descriptions" quite contradict with "describe_rule". Consider renaming "descriptions".

"describe_rule" function can be `functools.partial`-ed to not require `feature_extractor`. `feature_extractor` is not needed anywhere else so I think it will be nicer this way.

lookout/style/format/analyzer_report_template.jinja2
```diff
@@ -0,0 +1,23 @@
...
+{% if config.report_code_lines %}
+```{{ lang }}
+{% for l in range([0, line_number - 2]|max, [line_number + 1, code_lines|length]|min ) %}
+{{ "%d|%s" % (l, code_lines[l]) }}
+{% endfor %}
+```
+{% endif %}
+{% if new_code_line != code_lines[line_number - 1] %}
+```suggestion
+{{ new_code_line }}
```
vmarkovtsev: I am not sure but aren't we supposed to include the old line too here?

zurk:
addressed

lookout/style/format/analyzer.py
```diff
@@ -293,6 +243,122 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_service: D
...
+            if not line_ys or vnode_y.start.line == line_vnodes_y[0].start.line:
+                # collect all nodes on the same line
+                line_ys.append(yi)
+                line_ys_pred.append(y_predi)
+                line_vnodes_y.append(vnode_y)
+                line_winners.append(winner)
+                if yi != y_predi:
+                    generate_comment = True
+                continue
+            else:
```
vmarkovtsev: else + if -> elif

lookout/style/format/analyzer.py
```diff
@@ -145,58 +138,15 @@ def group_line_nodes(y, y_pred, vnodes_y, rule_winners):
...
+                if feature_extractor_output is None:
                     if self.config["report_parse_failures"]:
-                        comment = Comment()
-                        comment.file = file.path
-                        comment.confidence = 100
-                        comment.line = 1
-                        comment.text = "Failed to parse this file"
-                        comment.append(comment)
+                        comments.append(
+                            generate_comment(file.path, 100, 1, "Failed to parse this file"))
```
vmarkovtsev: This comment should be put on line #0 - this is a special case which is properly handled by Lookout

lookout/style/format/utils.py
```diff
@@ -142,7 +142,7 @@ def generate_comment(filename: str, confidence: int, line: int, text: str) -> Co
     assert 0 <= confidence <= 100, "Confidence should be in range 0~100 but value is '%s'" % \
                                    confidence
     assert isinstance(line, int), "Line should be integer but it's type is '%s'" % type(line)
-    assert 0 <= line, "Expected value >= 0 but got '%s'" % line
+    assert 0 <= line, "Expected value > 0 but got '%s'" % line
```
vmarkovtsev: Hence this should be reverted

lookout/style/format/analyzer.py
zurk: there is lines under `else`, but  not in `if`: https://github.com/src-d/style-analyzer/pull/335/files/78858ed7fa0dfd36046531fe0adefe92e020d0dc#diff-b71e0e71a29c83c6fa2ee406a76c3baeR358

zurk:
done

zurk:
there is a strange mistake in Python 3.5 3.7. 
I am figuring out.

zurk:
The fix is here: https://github.com/src-d/style-analyzer/issues/355 but the merge is blocked because it conflicts with Hugo refactoring and should be discussed.

vmarkovtsev:
@zurk I updated this PR with my edits. I am sorry that I did have time to finish the test - we need a `FeatureExtractor` and a trained `model` which I set tot None.

zurk:
ok

lookout/style/format/analyzer.py
```diff
@@ -27,6 +25,9 @@
 from lookout.style.format.rules import Rules, TrainableRules
 from lookout.style.format.utils import generate_comment, merge_dicts
 from lookout.style.format.virtual_node import VirtualNode
+import numpy
```
vmarkovtsev: Wow this is a bug in PyCharm import ordering

zurk: I'll fix it. (import order, not the bug in PyCharm :) )

lookout/style/format/analyzer.py
```diff
@@ -290,6 +240,159 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_service: D
...
+                code_lines=code_lines,             # original file code lines
+                new_code_line=new_code_line,       # code line suggested by our model
+                rules=rules.rules,                 # list of rules belonging to the model
+                winners=winners,                   # winner rule indices
+                vnodes=vnodes,                     # corrected VirtualNode-s
+                fixed_labels=fixed_labels,         # predicted labels for those nodes
+                confidence=confidence,             # overall confidence in the prediction, 0-100
+                describe_rule=_describe_rule,      # function to format a rule as text
+                describe_change=_describe_change,  # function to format a change as text
+                zip=zip,                           # Jinja2 does not have zip() by default
```
vmarkovtsev: Should be removed if we don't use in the template. I guess it didn't work? Does it work as `zip(a, b, c)`?

zurk: oh, wait a second...

zurk:
now it should be fine and @EgorBu you can use `FormatAnalyzer._analyze_unrendered()` as I do to avoid parsing logs after we merge this PR.

Issue #340, opened 2018-11-22
Add feature to blaklist rules based on support and confidence thresholds


warenlg:
As discussed with the team during 21/11/18 sync, rules should be filtered at the `analyze()` step based on given support and confidence thresholds.

PR #342, opened 2018-11-22
Filter rules according to confidence and support thresholds

warenlg:

Fixed https://github.com/src-d/style-analyzer/issues/340


lookout/style/format/analyzer.py
```diff
@@ -129,6 +131,8 @@ def group_line_nodes(y, y_pred, vnodes_y, rule_winners):
                             len(head_files), lang, lang)
                 continue
             rules = self.model[lang]
+            rules.filter_by_relevance(self.config["confidence_threshold"],
```
vmarkovtsev: Let's split this into two calls:

```python
rules = rules.filter_by_confidence().filter_by_support()
```

lookout/style/format/rules.py
```diff
@@ -142,6 +142,22 @@ def predict(self, X: numpy.ndarray, vnodes_y: Sequence[VirtualNode],
...
+        :param support_threshold: Minimum support value.
+        :return: `Rules` class with filtered rules.
+        """
+        self._log.debug("Filtering rules with confidence and support higher than %d and %d",
+                        confidence_threshold, support_threshold)
+        rules = [rule for rule in self._rules if rule.stats.conf > confidence_threshold
+                 and rule.stats.support > support_threshold]
+        self._log.debug("Number of rules decreased from %d to %d", len(self._rules), len(rules))
+        self._rules = Rules(rules, self._origin_config)
+        return self
```
vmarkovtsev: We should create a new instance of `Rules` rather than mutate the current one

PR #363, opened 2018-11-30
Resolve the conflict between src-d/ml and src-d/style-analyzer scikit-learn versions

m09:
style-analyzer was incompatible with sklearn 0.20 and src-d/ml recently bumped to it. This PR fixes the main incompatibility + the main warning we got after upgrade.


warenlg:
+1 on this PR, thanks Hugo

Issue #364, opened 2018-12-02
Value error in filter_uast_breaking_preds


EgorBu:
Hi,
I found that `filter_uast_breaking_preds` doesn't handle one type of errors (that is somehow handled in other cases):
```
ERROR:53a5:EventListener:FAIL ?
Traceback (most recent call last):
  File "/lookout-sdk-ml/lookout/core/event_listener.py", line 181, in wrapped_catch_them_all
    return func(self, request, context)
  File "/lookout-sdk-ml/lookout/core/event_listener.py", line 202, in wrapped_handle
    return getattr(self.handlers, method_name)(request)
  File "/lookout-sdk-ml/lookout/core/manager.py", line 100, in process_push_event
    model = analyzer.train(ptr, mycfg, self._data_service)
  File "/lookout-sdk-ml/lookout/core/data_requests.py", line 152, in wrapped_with_uasts_and_contents
    return func(cls, ptr, config, data_service, files=files, **data)
  File "/style-analyzer/lookout/style/format/quality_report.py", line 536, in train
    node_parents=node_parents, rule_winners=rule_winners, log=cls.log)
  File "/style-analyzer/lookout/style/format/postprocess.py", line 80, in filter_uast_breaking_preds
    stub, content_before[start:end], filename="", language=feature_extractor.language)
  File "/lookout-sdk-ml/lookout/core/data_requests.py", line 201, in parse_uast
    language=language)
ValueError: b'  <Menu.Item icon="edit" secondaryText="\xe2\x8c' has type str, but isn't valid UTF-8 encoding. Non-UTF-8 strings must be converted to unicode objects before being added.
```

Repo:
```
# format: "repository to-commit from-commit"
https://github.com/segmentio/evergreen ba22d511dad83c072842e47801ef42697d142f7c 1030eca5da38dce4e5047c23a3ea7fc0c246b8ce
```

m09:
We switched from BblfshClient to lookout and one expects binary while the other expectes utf-8 strings, that's why it reappeared I think.

PR #372, opened 2018-12-04
Fix ValueError for non UTF-8 chars and fix the number of filtered rules in the report

warenlg:
Fixed https://github.com/src-d/style-analyzer/issues/364 and the number of filtered rules in the report


lookout/style/format/quality_report_noisy.py
```diff
@@ -262,6 +262,8 @@ def quality_report_noisy(true_repo: str, noisy_repo: str, bblfsh: str, language:
     true_content = get_content_from_repo(true_repo)
     noisy_content = get_content_from_repo(noisy_repo)
     true_files, noisy_files, lines_changed = get_difflib_changes(true_content, noisy_content)
+    if not true_files:
+        raise ValueError("Noisy repo should count at leat one artificial mistake")
```
vmarkovtsev:
```suggestion
        raise ValueError("Noisy repo should count at least one artificial mistake")
```

PR #373, opened 2018-12-04
Add a cache and separate the function to check UAST

m09:
The speed up is not that significant sadly. The sheer number of calls to bblfsh is the main bottleneck.


lookout/style/format/postprocess.py
```diff
@@ -61,9 +65,39 @@ def filter_uast_breaking_preds(
     :return: List of predictions indices that are considered valid i.e. that are not breaking
              the UAST.
     """
+
+    def parse_partial_code(parent: bblfsh.Node, content: str
+                           ) -> Optional[Tuple[bblfsh.Node, int, int]]:
+        descendants = []
```
creachadair: Drive-by: There's quite a bit going on in this function and the nesting makes it difficult to understand the flow. Would you be willing to consider hoisting this to the top level and equipping it with a docstring?

m09: :+1:

lookout/style/format/postprocess.py
```diff
@@ -37,12 +37,62 @@ def check_uasts_are_equal(uast1: bblfsh.Node, uast2: bblfsh.Node) -> bool:
...
 
+def _parse_code(parent: bblfsh.Node, content: str, stub: "bblfsh.aliases.ProtocolServiceStub",
+                parsing_cache: MutableMapping[int, Optional[Tuple[bblfsh.Node, int, int]]],
+                language: str, node_parents: Mapping[int, bblfsh.Node], logger: Logger,
+                path: str
+                ) -> Optional[Tuple[bblfsh.Node, int, int]]:
+    """
+    Find a parent that Babelfish can parse and parse it.
+
+    Iterates on the parents of the current virtual node until it is parseable and return the parsed
```
vmarkovtsev:
```suggestion
    Iterates over the parents of the current virtual node until it is parseable and returns the parsed
```

lookout/style/format/postprocess.py
```diff
@@ -37,12 +37,62 @@ def check_uasts_are_equal(uast1: bblfsh.Node, uast2: bblfsh.Node) -> bool:
...
+    UAST or None if it reached the root without finding a parseable parent.
```
vmarkovtsev:
```suggestion
    UAST or None if it reaches the root without finding a parseable parent.
```

lookout/style/format/postprocess.py
```diff
@@ -37,12 +37,62 @@ def check_uasts_are_equal(uast1: bblfsh.Node, uast2: bblfsh.Node) -> bool:
     return True
 
 
+def _parse_code(parent: bblfsh.Node, content: str, stub: "bblfsh.aliases.ProtocolServiceStub",
+                parsing_cache: MutableMapping[int, Optional[Tuple[bblfsh.Node, int, int]]],
+                language: str, node_parents: Mapping[int, bblfsh.Node], logger: Logger,
+                path: str
+                ) -> Optional[Tuple[bblfsh.Node, int, int]]:
+    """
+    Find a parent that Babelfish can parse and parse it.
```
vmarkovtsev:
```suggestion
    Find a parent node that Babelfish can parse and parse it.
```

lookout/style/format/postprocess.py
```diff
@@ -37,12 +37,62 @@ def check_uasts_are_equal(uast1: bblfsh.Node, uast2: bblfsh.Node) -> bool:
...
+
+
 def filter_uast_breaking_preds(
         y: numpy.ndarray, y_pred: numpy.ndarray, vnodes_y: Sequence[VirtualNode],
         vnodes: Sequence[VirtualNode], files: Mapping[str, File],
         feature_extractor: FeatureExtractor, stub: "bblfsh.aliases.ProtocolServiceStub",
-        vnode_parents: Mapping[int, bblfsh.Node], node_parents: Mapping[str, bblfsh.Node],
-        rule_winners: numpy.ndarray, log: logging.Logger) -> Iterable[int]:
+        vnode_parents: Mapping[int, bblfsh.Node], node_parents: Mapping[int, bblfsh.Node],
+        rule_winners: numpy.ndarray, log: Logger) -> Tuple[numpy.ndarray,
```
vmarkovtsev:
```suggestion
        rule_winners: numpy.ndarray, log: Logger
        ) -> Tuple[numpy.ndarray, numpy.ndarray, Sequence[VirtualNode], numpy.ndarray, Sequence[int]]

```

vmarkovtsev:
Overall :+1: if this works :) Do we have any tests (not specific to the cache but rather generic white box)?

m09:
We have some tests but they are annoying to maintain currently (they require a trained model which makes many changes ---even correct ones--- fail the test. I'll add a test that is not model-dependent.

For this PR in particular I also checked in addition to running the tests that it was consistent in the visualizer before pushing.

m09:
@vmarkovtsev ready for next round. I removed the old test that was model dependent and added one that just tests the logic of the uast check.

warenlg:
@m09 Thank you for the PR and the tests. I just tested the logic and the number of predictions filtered out because breaking the UAST remains the same (43887 selected out of 45702 before and after your PR).

As regards the speed improvement, we pass from 448s to 406s user, but this includes the feature extraction and other steps, and I evaluate the speed gain of the postprocessing phase around x1.5, thanks !

m09:
Thanks @warenlg for the in-depth tests :+1:

Issue #375, opened 2018-12-05
Rule number increased during review event


zurk:
Just noted this strange behavior it in my current work. It can be related to my changes.
So it is just a reminder to have a look closer lately.

m09:
could it just be `insert_noops: True`? @EgorBu made the same remark after the change.

zurk:
Looks like a found a problem. See the related PR

PR #379, opened 2018-12-05
Fix artificial rules leak

zurk:
Fixes https://github.com/src-d/style-analyzer/issues/375
Also, fix Rule filtering. It was not applied. 
Fix strange behavior when analyzer tries to replace everything with `␣'`

Improve code a little bit.

### DIFF
--- a/lookout/style/format/research/seq_matching_vis.py
+++ b/lookout/style/format/research/seq_matching_vis.py
@@ -50,7 +50,7 @@ def visualize(input_filename: str, bblfsh: str, language: str, model_path: str) 
     fe = FeatureExtractor(language=language, **rules.origin_config["feature_extractor"])
     X, y, vnodes_y, vnodes = fe.extract_features([file])
 
-    y_pred, _ = rules.predict(X, vnodes_y, vnodes, fe)
+    y_pred, _, _ = rules.predict(X, vnodes_y, vnodes, fe)
 
     # collect lines with mispredictions - could be removed
     mispred_lines = set()
--- a/lookout/style/format/rules.py
+++ b/lookout/style/format/rules.py
@@ -5,7 +5,7 @@ import functools
 from importlib import import_module
 from itertools import islice
 import logging
-from typing import (Any, Dict, Iterable, List, Mapping, MutableMapping, NamedTuple, Optional,
+from typing import (Any, Dict, Iterable, List, Mapping, NamedTuple, Optional,
                     Sequence, Set, Tuple, Union)
 
 from bblfsh import role_name
@@ -69,7 +69,7 @@ class Rules:
         """
         super().__init__()
         assert rules is not None, "rules may not be None"
-        self._rules = rules
+        self._rules = tuple(rules)  # Rule list is constant
         self._compiled = self._compile(rules)
         self._origin_config = origin_config
 
@@ -116,10 +116,8 @@ class Rules:
         return prediction
 
     def predict(self, X: numpy.ndarray, vnodes_y: Sequence[VirtualNode],
-                vnodes: Sequence[VirtualNode], feature_extractor: FeatureExtractor,
-                return_originals: bool = False
-                ) -> Union[Tuple[numpy.ndarray, numpy.ndarray, numpy.ndarray, numpy.ndarray],
-                           Tuple[numpy.ndarray, numpy.ndarray]]:
+                vnodes: Sequence[VirtualNode], feature_extractor: FeatureExtractor
+                ) -> Tuple[numpy.ndarray, numpy.ndarray, "Rules"]:
         """
         Predict classes given the input features and metadata.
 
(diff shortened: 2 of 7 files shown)


lookout/style/format/research/seq_matching_vis.py
```diff
@@ -50,7 +50,8 @@ def visualize(input_filename: str, bblfsh: str, language: str, model_path: str)
     fe = FeatureExtractor(language=language, **rules.origin_config["feature_extractor"])
     X, y, vnodes_y, vnodes = fe.extract_features([file])
 
-    y_pred, _ = rules.predict(X, vnodes_y, vnodes, fe)
+    # TODO: fix rule visualization if new rules were added
+    y_pred, _, new_rules = rules.predict(X, vnodes_y, vnodes, fe)
```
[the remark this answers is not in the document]
m09: As far as I know this code as not been used. @EgorBu used it to experiment for a few days when we were building the evaluation datasets but dropped it if I recall correctly. It's not used in visualization at least.

lookout/style/format/rules.py
```diff
@@ -173,48 +166,43 @@ def _get_composite(feature_extractor: FeatureExtractor, labels: Tuple[int, ...])
...
         """
         quotes_classes = {CLASS_INDEX[CLS_DOUBLE_QUOTE], CLASS_INDEX[CLS_SINGLE_QUOTE]}
+        processed_rules = list(self.rules)
         processed_y = y_pred.copy()
         processed_winners = winners.copy()
-        y_indices = {id(vnode): i for i, vnode in enumerate(vnodes_y)}
         new_rules = {}
+
+        def append_new_rule(labels: Tuple[int, ...], y_i: int, conf: float, support: int) -> None:
+            if labels in new_rules:
```
m09: this does not assign the correct stats if a previous rule on the same labels was created with different stats.

lookout/style/format/rules.py
```diff
@@ -173,48 +166,43 @@ def _get_composite(feature_extractor: FeatureExtractor, labels: Tuple[int, ...])
...
-        rule = self._get_new_rule(feature_extractor, tuple(labels), new_rules)
-        winners[y_i] = rule
-        y[y_i] = self._rules[rule].stats.cls
-
-    def harmonize_quotes(self, X: numpy.ndarray, y_pred: numpy.ndarray,
+    def harmonize_quotes(self, y_pred: numpy.ndarray,
                          vnodes_y: Sequence[VirtualNode], vnodes: Sequence[VirtualNode],
                          winners: numpy.ndarray, feature_extractor: FeatureExtractor
-                         ) -> Tuple[numpy.ndarray, numpy.ndarray]:
+                         ) -> Tuple[numpy.ndarray, numpy.ndarray, List[Rule]]:
```
m09: We should return a complete `Rules` instance

lookout/style/format/visualizer/server.py
```diff
@@ -106,7 +106,7 @@ def return_features() -> Response:
                     "ground_truths": y.tolist(),
                     "predictions": y_pred.tolist(),
                     "sibling_indices": sibling_indices,
-                    "rules": describe_rules(rules.rules, fe),
+                    "rules": describe_rules(new_rules, fe),
```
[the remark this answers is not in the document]
m09: yup! When the function will return a `Rules` instance we can even just override rules in the rules.predict call and not touch this line :)

lookout/style/format/research/seq_matching_vis.py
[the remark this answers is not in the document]
zurk: wow, I did not pay attention it is in research directory. 😄

lookout/style/format/rules.py
zurk: Yeah. do you think it is a good idea to create a new rule here too?

m09: yes sadly it is necessary to have as many rules as (cls, precision, tuple) comboes :(

zurk: So there is a case when we can reuse rule. They goes from 
```python
            if labels1[-1] not in quotes_classes or labels3[0] not in quotes_classes:
                append_new_rule(vnode1.y, y_i_1, 1., 1)
                append_new_rule(vnode3.y, y_i_3, 1., 1)
```

so it is better to keep the logic, but chage `new_rules` indexes.  that what I did

lookout/style/format/visualizer/server.py
```diff
@@ -106,7 +106,7 @@ def return_features() -> Response:
                     "ground_truths": y.tolist(),
                     "predictions": y_pred.tolist(),
                     "sibling_indices": sibling_indices,
-                    "rules": describe_rules(rules.rules, fe),
+                    "rules": describe_rules(rules, fe),
```
m09: this needs to be changed back to `rules.rules` now that `predict` returns a `Rules` and not `Sequence[Rule]`

m09: just noticed Vadim also made a comment on that line, sorry for the duplicate

zurk:
should be good now

PR #383, opened 2018-12-07
Improve the noisy quality report rendering

warenlg:

This is an example of the current rendering
pr_curve_jquery_src_s80


lookout/style/format/cmdline_tools.py
```diff
@@ -73,7 +73,7 @@ def add_rules_thresholds(my_parser: ArgumentParser):
 
     :param my_parser: Parser to add the arguments to.
     """
-    my_parser.add_argument("--confidence-threshold", type=float, default=0.95,
+    my_parser.add_argument("--conf-threshold", type=float, default=0.95,
```
vmarkovtsev: "conf" can stand for "config", too, so better to leave the old name.

m09: we used conf in the Rule NamedTuple though

vmarkovtsev: Yes, given the Rule context that's fine. Here we've got command line arguments.

PR #384, opened 2018-12-07
Add marker for artificial rules and improve viz

### LINKED ISSUE #382: Add artificial marker to rules to be able to distinguish between normal and artificial rules.

m09:
Fixes #382.


lookout/style/format/descriptions.py
```diff
@@ -104,8 +118,8 @@ def describe_sample_ordinal(feature: OrdinalFeature, values: ndarray) -> str:
 
 
 @singledispatch
-def describe_rule_splits(feature: BagFeature, name: str,
-                         splits: List[Tuple[bool, floating, int]]) -> str:
+def _describe_rule_splits(feature: BagFeature, name: str,
```
vmarkovtsev: We should move it to FE class eventually because it is tightly coupled and I don't like playing C 😉

Issue #387, opened 2018-12-07
Disable label filtering on analyze stage


zurk:
Now it filters everything. 

log example:

```
DEBUG:FormatAnalyzer:Analyze tasks/compile-lockfile.js
DEBUG:FeaturesExtractor:Parsed 447 vnodes
DEBUG:FeaturesExtractor:Removed 9/10 labels by support 80
DEBUG:FeaturesExtractor:Features shape: (224, 500)
DEBUG:FeaturesExtractor:Labels shape: (224,)
DEBUG:Rules:predicting 224 samples using 96 rules
DEBUG:Rules:compiling 99 rules
INFO:FormatAnalyzer:Non UAST breaking predictions: 151 selected out of 224
DEBUG:FormatAnalyzer:Analyze tasks/cra.js
DEBUG:FeaturesExtractor:Parsed 1503 vnodes
DEBUG:FeaturesExtractor:Removed 10/12 labels by support 80
DEBUG:FeaturesExtractor:Features shape: (752, 500)
DEBUG:FeaturesExtractor:Labels shape: (752,)
DEBUG:Rules:predicting 752 samples using 96 rules
DEBUG:Rules:compiling 100 rules
INFO:FormatAnalyzer:Non UAST breaking predictions: 569 selected out of 752
DEBUG:FormatAnalyzer:Analyze tasks/screencast-start.js
DEBUG:FeaturesExtractor:Parsed 619 vnodes
DEBUG:FeaturesExtractor:Removed 8/9 labels by support 80
DEBUG:FeaturesExtractor:Features shape: (310, 500)
DEBUG:FeaturesExtractor:Labels shape: (310,)
DEBUG:Rules:predicting 310 samples using 96 rules
```

zurk:
So, the problem is that feature extraction is different when you train the model and when you apply it.
So it is clear that `FeatureExtractor` is Transformer and there are `fit()` and `transform()` methods should be implemented. For now, I just apply an ad-hoc solution to fix the problem.

zurk:
As soon as a bug is fixed I close this issue and create another one about `FeatureExtractor` refactoring.

PR #390, opened 2018-12-08
Disable label filtering on analyze stage

zurk:
temporary `fix` for https://github.com/src-d/style-analyzer/issues/387
Feature extractor should be refactored after the release.


lookout/style/format/analyzer.py
```diff
@@ -256,6 +256,7 @@ def generate_fixes(self, data_service: DataService, changes: Sequence) -> Iterab
                         find_deleted_lines(prev_file, file),
                     )))
                 fe = FeatureExtractor(language=lang, **rules.origin_config["feature_extractor"])
+                fe.cutoff_label_support = 0  # ad-hoc for https://github.com/src-d/style-analyzer/issues/387  # noqa: E501
```
m09: sorry for the nitpick but here we go: we should put comment line above to avoid noqa

Issue #388, opened 2018-12-08
Comment confidence calculation: IndexError: tuple index out of range


zurk:
The bug appears after https://github.com/src-d/style-analyzer/pull/379
It is easy to fix.

Traceback example: 
```
ERROR:EventListener:FAIL ?
Traceback (most recent call last):
  File "/lookout-sdk-ml/lookout/core/event_listener.py", line 181, in wrapped_catch_them_all
    return func(self, request, context)
  File "/lookout-sdk-ml/lookout/core/event_listener.py", line 202, in wrapped_handle
    return getattr(self.handlers, method_name)(request)
  File "/lookout-sdk-ml/lookout/core/manager.py", line 81, in process_review_event
    base_ptr, head_ptr, self._data_service)
  File "/style-analyzer/lookout/style/format/quality_report.py", line 228, in analyze
    [Changes(File(), f) for f in list(files)]):
  File "/style-analyzer/lookout/style/format/analyzer.py", line 271, in generate_fixes
    data_service.get_bblfsh(), rules)
  File "/style-analyzer/lookout/style/format/analyzer.py", line 323, in _generate_comments_data
    confidence = self._get_comment_confidence(line_ys, line_ys_pred, line_winners, rules)
  File "/style-analyzer/lookout/style/format/analyzer.py", line 347, in _get_comment_confidence
    confidence += rules.rules[winner].stats.conf
IndexError: tuple index out of range
```

PR #392, opened 2018-12-08
Fix comment confidence calculation

zurk:
Fixes https://github.com/src-d/style-analyzer/issues/388


zurk:
Wow, somehow not all changes are here. This one does nothing.

PR #397, opened 2018-12-08
Add noisy quality report on jquery repo

warenlg:


zurk:
Note: We still don't have the script. As we agreed it should be created in Makefile.

vmarkovtsev:
This means @warenlg should create the target in Makefile 😉

Issue #406, opened 2018-12-09
Filter composite labels by support does not work


zurk:
Despite the fact that we call `_remove_labels_with_low_support()` function during training:

https://github.com/src-d/style-analyzer/blob/abbe0f881b993139ac145dff54a4bb8476471397/lookout/style/format/feature_extractor.py#L266-L267

No filtering happens.

zurk:
This bug covers two another:
So there are two PRs:
https://github.com/src-d/style-analyzer/pull/407
https://github.com/src-d/style-analyzer/pull/408

PR #407, opened 2018-12-09
Make _remove_labels_with_low_support work

zurk:
Fixes https://github.com/src-d/style-analyzer/issues/406

`_remove_labels_with_low_support` tried to use generator twice. So second time was absolutely useless and no actual filtering was applied.

### DIFF
--- a/lookout/style/format/feature_extractor.py
+++ b/lookout/style/format/feature_extractor.py
@@ -1,7 +1,7 @@
 """Feature extraction module."""
 from collections import defaultdict, OrderedDict
 import importlib
-from itertools import chain, islice, zip_longest
+from itertools import islice, zip_longest
 import logging
 from typing import (Dict, Iterable, List, Mapping, Optional, Sequence, Tuple, Union)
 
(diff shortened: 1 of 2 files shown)


lookout/style/format/feature_extractor.py
```diff
@@ -264,7 +264,7 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None) \
         # filter composite labels by support
         if index_labels and self.cutoff_label_support > 0:
             self._remove_labels_with_low_support(
-                chain.from_iterable(vn for vn, _, _ in parsed_files))
+                list(chain.from_iterable(vn for vn, _, _ in parsed_files)))
```
m09: `[vn for vns, _, _ in parsed_files for vn in vns]`

lookout/style/format/feature_extractor.py
```diff
@@ -728,7 +728,7 @@ def _parse_file(self, contents: str, root: bblfsh.Node, path: str) -> \
             pos = node.end_position.offset
         return result, parents
 
-    def _remove_labels_with_low_support(self, vnodes: Iterable[VirtualNode]):
+    def _remove_labels_with_low_support(self, vnodes: List[VirtualNode]):
```
m09: `Sequence[VirtualNode]` and return type `None`

zurk:
Addressed  👍

lookout/style/format/tests/test_postprocess.py
```diff
@@ -23,6 +23,7 @@ def setUpClass(cls):
         cls.fe = FeatureExtractor(
             language=cls.language,
             **FormatAnalyzer._load_train_config({})[cls.language]["feature_extractor"])
+        cls.fe.cutoff_label_support = 0
```
m09: Missed this one sorry: it's better to change the config than modify the fe attribute

zurk: it is in tests. let's keep it as is. 🙏

m09: fixing tests because they're not testing the behavior but use internal implementation details is tech debt and we already have lots of debt here

m09:
@zurk: Travis fails because of a style issue

PR #408, opened 2018-12-09
Fix for CodeGenerator.gerenate() and FeatureExtractor._add_noops() if we remove labels with low support

zurk:
Depends on https://github.com/src-d/style-analyzer/pull/407

Both this methods relay that `y`-s in VirtualNode-s are the initial unfiltered targets. As soon as PR #407 make this filtering work we should fix these methods. 

Bug example for `FeatureExtractor._add_noops()`:
1. We filter `y` with value `⏎⏎⏎⏎⏎⏎⏎⏎⏎` and set it to `None`. `FeatureExtractor._add_noops()` see that `y=None` and adds NOOPs from both sides. 

Bug example for `CodeGenerator.gerenate()`:
2. We filter `y` with value `⏎⏎⏎␣⁻␣⁻␣⁻` and set `y=None`. Now we lose information about indentation change. Also `FeatureExtractor._add_noops()` inserts NOOPs between such VirtualNode and accumulated indentation. Both things confuse `CodeGenerator`.

So the main idea of PR is to add a new field to `VirtualNodes` called `y_original`. We set this field when any filtering is applied. 

PS: The fix is the ugliest one. And it is worth to include `y_original` to `VirtualNode.__init__()` feel free to ask me to do that if you have the same opinion.

### DIFF
--- a/lookout/style/format/code_generator.py
+++ b/lookout/style/format/code_generator.py
@@ -72,10 +72,12 @@ class CodeGenerator:
         """
         result = []
         for vnode, y_new in self._iterate_vnodes(vnodes, vnodes_y, y_pred):
-            vnode = vnode.copy()
-            if y_new != vnode.y:
-                vnode.y, vnode.y_old = y_new, vnode.y
-            result.append(vnode)
+            new_vnode = vnode.copy()
+            if hasattr(vnode, "y_original"):
+                new_vnode.y_original = vnode.y_original
+            if y_new != new_vnode.y:
+                new_vnode.y, new_vnode.y_old = tuple(y_new), new_vnode.y
+            result.append(new_vnode)
         return result
 
     def generate(self, vnodes: Sequence[VirtualNode], indentation: str) -> str:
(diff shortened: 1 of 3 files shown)


zurk:
My main worry about this PR is that Egor's evaluation report shows performance drop. I should check it with current mater to be 100% but if it is true, we should find a reason.

zurk:
Just a note that as soon as we merge https://github.com/src-d/style-analyzer/pull/407 and not this PR a lot of stuff in the master is not going to work as expected

cc @vmarkovtsev @m09 

Feel free to ask me about any explanations you need.

m09:
When we consider only a few lines of labeled nodes during `analyze` we do not remove `y` from all the other nodes. In the same vein here we should just add a test during the creation of `vnodes_y` that the class is in the supported labels and not touch their `y` attribute. This PR complicates an already unmaintainable part of the codebase so I don't feel comfortable with it.

zurk:
ok, I'd like this way of solving the problem. Let me try.

zurk:
I close this one and open new https://github.com/src-d/style-analyzer/pull/415 because it is completely different. Thanks, @m09 for a better solution!

PR #409, opened 2018-12-10
Speed up UAST breaking check

m09:


m09:
@warenlg measured the perf difference and it doesn't seem to be helping even though now we're not linear on `vnodes` anymore in the inner loop. We can merge anyway or close, as you prefer @vmarkovtsev.

PR #411, opened 2018-12-10
Fix string versions

### LINKED ISSUE #385: FomatModel saves version as string.
I am not sure if an issue should be here or in modelforge, but if you dump FormatAnalyzer you see
```
{'created_at': datetime.datetime(2018, 12, 7, 21, 51, 0, 861816),
<...>
 'version': ['1']}
```

However, modelforge expects this version to be int: https://github.com/src-d/modelforge/blob/master/modelforge/model.py#L248

vmarkovtsev:
Fixes #385


m09:
should also be fixed in https://github.com/src-d/lookout-sdk-ml/blob/master/lookout/core/analyzer.py#L103

PR #415, opened 2018-12-10
Fix for CodeGenerator.gerenate() and FeatureExtractor._add_noops() if we remove labels with low support

zurk:
Second attempt to fix missing `vnode.y`

Issue description: https://github.com/src-d/style-analyzer/pull/408#issue-237166170
and related issue: https://github.com/src-d/style-analyzer/issues/406

### DIFF
--- a/lookout/style/format/feature_extractor.py
+++ b/lookout/style/format/feature_extractor.py
@@ -1,7 +1,7 @@
 """Feature extraction module."""
 from collections import defaultdict, OrderedDict
 import importlib
-from itertools import islice, zip_longest
+from itertools import chain, islice, zip_longest
 import logging
 from typing import (Dict, Iterable, List, Mapping, Optional, Sequence, Tuple, Union)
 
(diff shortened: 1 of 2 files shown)


zurk:
`update_mappings` is called from two places, so it is better to keep it.

m09:
sorry I should have been clearer: I think we can centralize what we do in update_mappings in _remove_labels_with_low_support (and rename it _extract_labels).

zurk:
let's name it _extract_labels_mappings to be more precise?

m09:
the mappings are implementation details, the goal is to extract labels. But up to you it's really minor.

zurk:
just `_extract_labels` confuses me and I expect some kind of output with labels from a call like `_extract_labels(vnodes)`.

m09:
sure but I guess there is the same problem with _extract_labels_mappings. So maybe _compute_labels ? Anyway, any name will do really :)

lookout/style/format/feature_extractor.py
```diff
@@ -261,14 +261,18 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None) \
...
             self._fill_vnode_parents(file_parents, file_vnodes, uast, vnode_parents)
 
-        self._log.debug("Parsed %d vnodes", sum(len(vn) for vn, _, _ in parsed_files))
+        vnodes_parsed_number = sum(len(vn) for vn, _, _ in parsed_files)
+        self._log.debug("Parsed %d vnodes", vnodes_parsed_number)
         # filter composite labels by support
-        if index_labels and self.cutoff_label_support > 0:
-            self._remove_labels_with_low_support([vn for vns, _, _ in parsed_files for vn in vns])
+        if index_labels:
+            self._compute_labels_mappings([vn for vns, _, _ in parsed_files for vn in vns])
```
m09: we can go back to iterator :)

lookout/style/format/feature_extractor.py
```diff
@@ -728,21 +717,30 @@ def _parse_file(self, contents: str, root: bblfsh.Node, path: str) -> \
...
             if vnode.y is not None:
-                label_support[vnode.y] += 1
-        unsupported = {k for k, v in label_support.items() if v < self.cutoff_label_support}
-        for vnode in vnodes:
-            if vnode.y in unsupported:
-                vnode.y = None
-        self._log.debug("Removed %d/%d labels by support %d", len(unsupported), len(label_support),
+                support[vnode.y] += 1
+
+        for class_seq in sorted(support, key=lambda l: -support[l]):
```
m09: O(n log n) instead of O(n) and less readable

zurk:
Are you fine if I ask Vadim to review and merge it?

m09:
sure

lookout/style/format/feature_extractor.py
m09: if it was not an optimization but a feature it's fine though

zurk: It is small array ~100 of items, so does not really affect performance.
I just think it is a good idea to have this order: from the most frequent label to the least frequent.

m09: yup I agree

lookout/style/format/feature_extractor.py
```diff
@@ -728,21 +717,30 @@ def _parse_file(self, contents: str, root: bblfsh.Node, path: str) -> \
...
```
vmarkovtsev:
```python
self.labels_to_class_sequences = [key for key, val in support.items() if val >= self.cutoff_label_support]
```

m09: @vmarkovtsev: as discussed above sorting the labels by support makes sense to identify quickly if the label at hand is common or not

lookout/style/format/feature_extractor.py
m09: The for loop does nothing

PR #418, opened 2018-12-11
Clean _compute_labels_mappings and fix viz

m09:


lookout/style/format/feature_extractor.py
```diff
@@ -732,11 +733,9 @@ def _compute_labels_mappings(self, vnodes: Iterable[VirtualNode]) -> None:
...
                 support[vnode.y] += 1
 
-        for class_seq in sorted(support, key=lambda l: -support[l]):
-            # The idea of sorting is to have labels arranged from the most frequent to the less
-            if support[class_seq] < self.cutoff_label_support:
-                break
-        self.labels_to_class_sequences = [key for key, val in support.items()
+        self.labels_to_class_sequences = [key for key, val in sorted(support.items(),
+                                                                     key=itemgetter(1),
+                                                                     reverse=True)
```
zurk: yeah, a mess happened here. thank you👍 
Just let keep the comment and make it not so right shifted?

```python
        # The idea of sorting is to have labels arranged from the most frequent to the less
        self.labels_to_class_sequences = [
            key for key, val in sorted(support.items(), key=itemgetter(1), reverse=True)
            if val >= self.cutoff_label_support]
```

lookout/style/format/visualizer/server.py
```diff
@@ -42,20 +44,23 @@ def _rules_to_jsonable(rules: Rules, feature_extractor: FeatureExtractor
...
                   "line": int(vnode.start.line)},
         "end": {"offset": int(vnode.end.offset),
                 "col": int(vnode.end.col),
                 "line": int(vnode.end.line)},
         "value": vnode.value,
-        "path": vnode.path,
-        "roles": [role for role in vnode.node.roles] if vnode.node else [],
-        "y": vnode.y,
-        "internal_type": vnode.node.internal_type if vnode.node else None,
+        "path": vnode.path
```
zurk:
```suggestion
        "path": vnode.path,
```

But as you wish

m09: I agree. We should probably add that to flake8. I'll create an issue

PR #420, opened 2018-12-11
Fix flaky test in typos

m09:
The logic in `__eq__` is still broken but the test is not flaky anymore (by broken I mean that some logic should clearly not be in this class).

### DIFF
--- a/lookout/style/typos/tests/test_generation.py
+++ b/lookout/style/typos/tests/test_generation.py
@@ -41,6 +41,7 @@ class CandidatesSplitTest(unittest.TestCase):
 
 
 class GeneratorTest(unittest.TestCase):
+    @unittest.skip("CandidatesGenerator.__eq__ needs refactoring. Test is currently flaky.")
     def test_save_load(self):
         generator = CandidatesGenerator()
         generator.construct(VOCABULARY_FILE, VOCABULARY_FILE, FASTTEXT_DUMP_FILE,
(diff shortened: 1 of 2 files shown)


lookout/style/typos/tests/test_generation.py
```diff
@@ -50,7 +50,6 @@ def test_save_load(self):
             print(buffer.tell())
             buffer.seek(0)
             generator2 = CandidatesGenerator().load(buffer)
-
```
vmarkovtsev: We still need to mark this test as `@unittest.expectedFailure`. I investigated the problem and it is not easily fixed by those changes in `lookout/style/typos/generation.py`. The core root is I think because sometimes it compares a numpy array to an ASDF proto-array which casts itself to a numpy array and is not an instance of ndarray. So this is the fault of the test after all.

m09:
There were new tests failing so I only tagged the flaky tests that I could find ATM. This'll need a more thorough look.

I didn't use expectedFailure but skip since we don't really expect them to fail (they are flaky) so skip is more appropriate.

vmarkovtsev:
@m09 I liked the way you changed the comparison code actually, I miss it and hope that you will return it back :cry:

m09:
@vmarkovtsev I'll create an issue to document the current problems with `__eq__` and to make sure that we don't forget to come back to it

PR #426, opened 2018-12-11
Turn on prune_attributes by default

zurk:


zurk:
So, only +50-100% to test execution time. :) 
I expect the same performance drop on a Quality report and it can take more than 24h.

vmarkovtsev:
@zurk @m09 We need to turn attributes pruning off individually in all the tests except one (to have it covered).

zurk:
Yes, there is a more abstract issue for it: https://github.com/src-d/style-analyzer/issues/432

m09:
Regarding specifically attributes pruning, see https://github.com/src-d/style-analyzer/pull/439#issue-237773135. I think we'll have to disable it everywhere for now.

Issue #368, opened 2018-12-04
Investigate undetected mistakes in smoke dataset


m09:
Smoke dataset shows that there are mostly undetected mistakes.

zurk:
@zurk ping

zurk:
So it is still happening.

PR #431, opened 2018-12-11
Fix condition for error check

zurk:
This was easy 
Fix https://github.com/src-d/style-analyzer/issues/368

### DIFF
--- a/lookout/style/format/benchmarks/evaluate_smoke.py
+++ b/lookout/style/format/benchmarks/evaluate_smoke.py
@@ -222,6 +222,7 @@ def calc_metrics(bad_style_code: str, correct_style_code: str, fe: Optional[Feat
     :return: A dictionary with losses and predicted code for "global" and "local" indentation \
              strategies.
     """
+    # TODO (zurk): make optional arguments non-optional.
     if (vnodes is None) ^ (fe is None):
         raise ValueError("vnodes and fe should be both None or not None.")
     losses = {}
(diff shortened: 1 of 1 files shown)


lookout/style/format/benchmarks/evaluate_smoke.py
```diff
@@ -153,7 +153,7 @@ def calc_aligned_metrics(bad_style_code: str, correct_style_code: str, generated
     return misdetection, undetected, detected_wrong_fix, detected_correct_fix
 
 
-def calc_metrics(bad_style_code: str, correct_style_code: str, fe: Optional[FeatureExtractor],
+def  calc_metrics(bad_style_code: str, correct_style_code: str, fe: Optional[FeatureExtractor],
```
vmarkovtsev: Double space here
```suggestion
def calc_metrics(bad_style_code: str, correct_style_code: str, fe: Optional[FeatureExtractor],
```

vmarkovtsev: @zurk Why is everything optional here? I prefer to force existing values everywhere unless we really need otherwise.

zurk: I agree, but it is not so easy. because you can have 0 fixes for some files. 
I keep a todo for the future.

vmarkovtsev: If you have zero fixes, you still pass a valid FeatureExtractor instead of None, don't you?

zurk: no, I just do not have anything in this case. because I take `FeatureExtractor` from `FixData` and if there is no `FixData` => no `FeatureExtractor`

vmarkovtsev:
@zurk OK I see. You need to split `calc_metrics` into two functions in the future, and it's totally OK to duplicate code in them. The second is for empty comments. Also force non-None FeatureExtractor in CodeGenerator.

zurk:
After release?

vmarkovtsev:
Yes

vmarkovtsev:
@zurk Could you please create an issue

PR #439, opened 2018-12-11
Restrict hyper-parameter search to base model

m09:
@vmarkovtsev some context on why I set `prune_attributes` to `False` again:

In #426, we had the code:

```python
            backup_prune_attributes = lang_config["trainable_rules"]["prune_attributes"]
            lang_config["trainable_rules"]["prune_attributes"] = False
            try:
                # do hyperparam search and store result in bscv
            finally:
                lang_config["trainable_rules"]["prune_attributes"] = backup_prune_attributes
            # some more code
            lang_config["trainable_rules"].update(bscv.best_params_)
            trainable_rules = TrainableRules(**lang_config["trainable_rules"],
                                             origin_config=lang_config)
```
And the `.update(bscv.best_params_)` overrides the restoration of the original value of `prune_attributes` which means it's effectively always `False` during training.

Since I didn't use this code in this PR it was actually turned on and hanged in Travis (too long to compute). I think we should therefore turn it off before we find a better solution.

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -4,11 +4,10 @@ from itertools import chain
 import logging
 import os
 from pprint import pformat
-import threading
 from typing import Any, Iterable, List, Mapping, NamedTuple, Sequence, Tuple
 
+import bblfsh  # noqa: F401
 from jinja2 import Template
-from lookout.core import slogging
 from lookout.core.analyzer import Analyzer, ReferencePointer
 from lookout.core.api.service_analyzer_pb2 import Comment
 from lookout.core.api.service_data_pb2 import File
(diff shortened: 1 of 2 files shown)


vmarkovtsev:
@m09 You swap two lines and get everything correct.

vmarkovtsev:
Edit 2: as for the tests, I asked to turn pruning off in our tests, which should be our top priority now.

m09:
@vmarkovtsev just to put it another way: last PR did not turn attributes pruning on and to turn it on we also need to fix tests, which is orthogonal to this PR. IMO we should merge this PR and have a separate PR to turn attributes pruning on correctly (which means disabling it in most tests). This PR is in no way incompatible to turning on attributes pruning.

m09:
Not related but there is currently a bug where even with a fixed `random_state`, reproducibility tests are failing but only 50% of the time (very weird). I'm investigating.

vmarkovtsev:
Got your point.

m09:
Bug is fixed

PR #440, opened 2018-12-11
Save only on > 0 rules and raise exception for empty models

m09:
I'm not really familiar with all the latest developments in the report analyzers so at this point I only raised exceptions when trying to access inexistant rules. There are probably better specific tweaks that could be applied.

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -223,7 +223,10 @@ class FormatAnalyzer(Analyzer):
                 "\n\t".join("%-55s %.5E" % (fe.feature_names[i], importances[i])
                             for i in numpy.argsort(-importances)[:25] if importances[i] > 1e-5))
             # throw away imprecise classes
-            model[language] = trainable_rules.rules
+            if trainable_rules.rules.rules:
+                model[language] = trainable_rules.rules
+            else:
+                cls._log.warning("Model for %s had 0 rules. Skipping." % language)
         cls._log.info("trained %s", model)
         return model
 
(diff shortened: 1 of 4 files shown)


lookout/style/format/analyzer.py
```diff
@@ -223,7 +223,10 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_service: D
                 "\n\t".join("%-55s %.5E" % (fe.feature_names[i], importances[i])
                             for i in numpy.argsort(-importances)[:25] if importances[i] > 1e-5))
             # throw away imprecise classes
-            model[language] = trainable_rules.rules
+            if trainable_rules.rules.rules:
+                model[language] = trainable_rules.rules
+            else:
+                cls._log.warning("model for %s had 0 rule. Skipping" % language)
```
zurk:
```suggestion
                cls._log.warning("Model have 0 rule for %s. Skipping." % language)
```
?

m09:
I don't mind which exception to use. `NotFittedError` is fine in the sense that the model is not fitted for this language, the others are also ok.

lookout/style/format/analyzer.py
m09: we have really inconsistent style so I think it's really arbitrary to pick one or the other (the 2 styles seem to be 50/50 in analyzer.py). We could do a pass to harmonize at some point though

vmarkovtsev:
```suggestion
                cls._log.warning("model for %s has 0 rules. Skipping" % language)
```

lookout/style/format/quality_report.py
```diff
@@ -107,6 +108,8 @@ def generate_model_report(model: FormatModel,
     languages = languages if isinstance(languages, Iterable) else [languages]
     template = _load_jinja2_template("templates/model_report.md.jinja2")
     for language in languages:
+        if language not in model:
```
vmarkovtsev: Not related but I have just noticed a typo upwards:
```
:param languages: **Lnguages** for which report
```

Issue #313, opened 2018-11-08
Create a script to output a single quality report gathering all specific reports we built


warenlg:
Currently, to evaluate the performance of the style-analyzer, we use 3 different quality reports:
* The one output by the `robust-eval` command that measure through precision and recall the analyzer's ability to fix artificial style mistakes in a repository.
* The one based on the smoke dataset, focusing on specific categories of style mistakes.
* The one Egor's worked on in this PR https://github.com/src-d/style-analyzer/pull/251

At the end we should run a single script that would compiled all those reports and output a final single performance report.

EgorBu:
It could be done if all reports will be done as analyzers. 
Example for quality report https://github.com/src-d/style-analyzer/pull/326

vmarkovtsev:
Ad-hoc command line entries are fine too, I think.

vmarkovtsev:
This will be a Makefile and the work happens in https://github.com/src-d/style-analyzer/pull/397 by @warenlg

PR #454, opened 2018-12-12
Add make release-report

zurk:
Closes https://github.com/src-d/style-analyzer/issues/313

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,14 +14,11 @@ $(MAKEFILE):
 	git clone --quiet --depth 1 -b $(CI_BRANCH) $(CI_REPOSITORY) $(CI_PATH);
 -include $(MAKEFILE)
 
-
-.PHONY: check
 check:
 	! grep -R /tmp lookout/style/*/tests
 	flake8 --count
 	pylint lookout
 
-.PHONY: docker-test
 docker-test:
 	docker ps | grep bblfshd  # bblfsh server should be run. Try `make bblfsh-start` command.
 	docker run --rm -it --network host \
@@ -30,10 +27,34 @@ docker-test:
 		--entrypoint python3 -w /style-analyzer \
 			srcd/style-analyzer:test -m unittest discover
 
-.PHONY: bblfsh-start
 bblfsh-start:
 	! docker ps | grep bblfshd # bblfsh server has been run already.
 	docker run -d --rm --name style_analyzer_bblfshd --privileged -p 9432\:9432 \
 		bblfsh/bblfshd\:v2.10.0
 	docker exec style_analyzer_bblfshd bblfshctl driver install \
 		javascript docker://bblfsh/javascript-driver\:v1.2.0
+
+REPORTS_DIR ?= reports
+REPORT_VERSION ?= untagged
+REPORT_DIR ?= $(REPORTS_DIR)/$(REPORT_VERSION)
+SMOKE_REPORT_DIR ?= $(REPORT_DIR)/js_smoke
+NOISY_REPORT_DIR ?= $(REPORT_DIR)/noisy
+QUALITY_REPORT_DIR ?= $(REPORT_DIR)/quality
+SMOKE_INIT ?= ./lookout/style/format/benchmarks/data/js_smoke_init.tar.xz
+
+$(SMOKE_REPORT_DIR) $(NOISY_REPORT_DIR) $(QUALITY_REPORT_DIR):
+	mkdir -p $@
+report-smoke: $(SMOKE_REPORT_DIR)
+	python3 -m lookout.style.format gen-smoke-dataset $(SMOKE_INIT) $(SMOKE_REPORT_DIR)/dataset \
+		2>&1 | tee $(SMOKE_REPORT_DIR)/dataset_logs.txt
+	python3 -m lookout.style.format eval-smoke-dataset $(SMOKE_REPORT_DIR)/dataset \
+		$(SMOKE_REPORT_DIR)/report 2>&1 | tee $(SMOKE_REPORT_DIR)/report_logs.txt
+report-noisy: $(NOISY_REPORT_DIR)
+	python3 -m lookout.style.format quality-report-noisy -o $(NOISY_REPORT_DIR) \
+		2>&1 | tee $(NOISY_REPORT_DIR)/logs.txt
+report-quality: $(QUALITY_REPORT_DIR)
+	python3 -m lookout.style.format.benchmarks.top_repos_quality -o $(QUALITY_REPORT_DIR) \
+		2>&1 | tee $(QUALITY_REPORT_DIR)/logs.txt
+report-release: report-smoke report-noisy report-quality
+
+.PHONY: check docker-test bblfsh-start report-smoke report-noisy report-quality report-release
--- a/lookout/style/format/cmdline_tools.py
+++ b/lookout/style/format/cmdline_tools.py
@@ -52,7 +52,9 @@ def add_rules_thresholds(my_parser: ArgumentParser):
 
     :param my_parser: Parser to add the arguments to.
     """
-    my_parser.add_argument("--confidence-threshold", type=float, default=0.95,
+    # This default values also used in `make report-noisy`.
+    # You should think twice before changing it.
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -38,3 +38,23 @@ bblfsh-start:
 	docker exec style_analyzer_bblfshd bblfshctl driver install \
 		javascript docker://bblfsh/javascript-driver\:v1.2.0
 	(docker exec -it style_analyzer_bblfshd bblfshctl driver list | grep javascript) || (docker rm -f style_analyzer_bblfshd && false)
+
+REPORTS_DIR = reports
```
vmarkovtsev: Please change all "=" to "?=" (this means set if not defined externally)

Makefile
```diff
@@ -38,3 +38,23 @@ bblfsh-start:
...
+REPORT_VERSION = untagged
+REPORT_DIR = $(REPORTS_DIR)/$(REPORT_VERSION)
+SMOKE_REPORT_DIR = $(REPORT_DIR)/js_smoke
+NOISY_REPORT_DIR = $(REPORT_DIR)/noisy
+QUALITY_REPORT_DIR = $(REPORT_DIR)/quality
+SMOKE_INIT = ./lookout/style/format/benchmarks/data/js_smoke_init.tar.xz
+.PHONY: release-report
+release-report:
```
vmarkovtsev:
```
$(SMOKE_REPORT_DIR) $(NOISY_REPORT_DIR) $(QUALITY_REPORT_DIR):
    mkdir -p $@
report_smoke: $(SMOKE_REPORT_DIR)
    # .. smoke code
noisy_report: $(NOISY_REPORT_DIR)
    # noisy code
...etc...
release-report: report_smoke noisy_report quality_report
```

zurk: I hope I get it right.
So no `.PHONY` for any of this commands?

m09: `.PHONY` is required for these commands.

Also should all be put in one single line to avoid bloating the Makefile

Makefile
```diff
@@ -37,3 +37,27 @@ bblfsh-start:
...
+
+$(SMOKE_REPORT_DIR) $(NOISY_REPORT_DIR) $(QUALITY_REPORT_DIR):
+	mkdir -p $@
+report-smoke: $(SMOKE_REPORT_DIR)
+	python3 -m lookout.style.format gen-smoke-dataset $(SMOKE_INIT) $(SMOKE_REPORT_DIR)/dataset \
+		2>&1 | tee $(SMOKE_REPORT_DIR)/dataset_logs.txt
+	python3 -m lookout.style.format eval-smoke-dataset $(SMOKE_REPORT_DIR)/dataset \
+		$(SMOKE_REPORT_DIR)/report 2>&1 | tee $(SMOKE_REPORT_DIR)/report_logs.txt
+report-noisy: $(NOISY_REPORT_DIR)
+	python3 -m lookout.style.format quality-report-noisy -o $(NOISY_REPORT_DIR) \
```
m09: It should be better not to specify arguments in Makefile but rather use default ones (and change them if required) to avoid having defaults out of sync everywhere

Makefile
```diff
@@ -37,3 +37,27 @@ bblfsh-start:
...
+		--confidence-threshold 0 --support-threshold 80 --precision-threshold 0.95 \
+		2>&1 | tee $(NOISY_REPORT_DIR)/logs.txt
+report-quality:
```
m09: should depend on QUALITY_REPORT_DIR

Makefile
zurk: there are no defaults and I think like they should not be there.

Makefile
zurk: sorry I do not get the last one.

Makefile
m09: WDYM? There are defaults (https://github.com/src-d/style-analyzer/blob/master/lookout/style/format/cmdline_tools.py#L55 for example)

m09: I was not speaking about arguments that are currently mandatory but rather the optional ones that currently have defaults that are overridden here, just to be clear

zurk: oh, did not see this one you find. I'd prefer just delete this default value.

Makefile
m09: basically, transforming all the lines
```
.PHONY: some-target
...
.PHONY: some-other-target
```
into
```
.PHONY: some-target some-other-target
```

Makefile
zurk: just to always have more explicit way of what exactly we are running

m09: it's fine to have default values in cmdline_tools.py, we just need to avoid having another source of truth here

Makefile
```diff
@@ -37,3 +37,28 @@ bblfsh-start:
...
+
+REPORTS_DIR ?= reports
+REPORT_VERSION ?= untagged
+REPORT_DIR ?= $(REPORTS_DIR)/$(REPORT_VERSION)
+SMOKE_REPORT_DIR ?= $(REPORT_DIR)/js_smoke
+NOISY_REPORT_DIR ?= $(REPORT_DIR)/noisy
+QUALITY_REPORT_DIR ?= $(REPORT_DIR)/quality
+SMOKE_INIT ?= ./lookout/style/format/benchmarks/data/js_smoke_init.tar.xz
+
+.PHONY: report-smoke report-noisy report-quality report-release
```
m09: should also include previous .PHONY targets (start of the file)

Makefile
m09: this is somewhat orthogonal to this PR though. That some arguments might be better off not having default is one thing, but as long as they do have a default we should not override them here to not have 2 sources of truth. We can create an issue to discuss the CLI but once again this is orthogonal to this PR.

lookout/style/format/benchmarks/top_repos_quality.py
```diff
@@ -25,28 +25,28 @@
...
-    "https://github.com/axios/axios 21ae22dbd3ae3d3a55d9efd4eead3dd7fb6d8e6e 138108ee56bd689305ae505a66b48d5e9c8aa494",  # noqa: E501
-    "https://github.com/GoogleChromeLabs/carlo b8ce2bca042c757b13fc82a3e059980342ddd9a8 26262aad740b7255f17950251ae344b4823572a4",  # noqa: E501
-    "https://github.com/nodejs/citgm 0c4c7ccdd1cad8ce9506e34ca523787ba18cafe2 d21e2a87aaa9e9f50c6175eddc54054a32c64a24",  # noqa: E501
-    "https://github.com/facebook/create-react-app 32106d216e4c31fda30ec475f9f03186d116c893 ffc63d55976f9cbcce7f33dc7c45b3c2190a5924",  # noqa: E501
-    "https://github.com/expressjs/express b4eb1f59d39d801d7365c86b04500f16faeb0b1c 56e90e3c7267782febe35754806ce3f63b527485",  # noqa: E501
-    "https://github.com/freeCodeCamp/freeCodeCamp cf65516cce60645a417e44c4fcea7418ca920572 c353c4c659c3dcb19524ba893f170805c931a44a",  # noqa: E501
-    "https://github.com/jquery/jquery dae5f3ce3d2df27873d01f0d9682f6a91ad66b87 d9a099a58e1bb1f158ea66ec55534770be442907",  # noqa: E501
+    # "https://github.com/30-seconds/30-seconds-of-code 3a122c9cfcbdc091227879a06a32bc67ccd0d35d c8c60895e80b8bc90583502accdaa339b794609c",  # noqa: E501
```
vmarkovtsev: Forgot to uncomment :D

Makefile
vmarkovtsev: I would say that the default values are default exactly because we think that we can provide a sane no-brainer. It makes sense to call the defaults those values which are used to generate the release report. I am totally fine with Hugo's proposal because we print all the configuration in the report. If we don't print some arguments, then we should fix our report.

lookout/style/format/benchmarks/top_repos_quality.py
zurk: oops, it is my local stuff, yeah

zurk:
Sorry for the mess if you saw it. I think it is done.

PR #473, opened 2018-12-13
Fix test_train_review_analyzer_integration

zurk:


vmarkovtsev:
Tests pass, Travis is having problems with reporting.

Issue #461, opened 2018-12-13
Code generation behaves weird


vmarkovtsev:
As Egor showed today, we predict correct fixes but generate wrong code suggestions: new lines are missed.

PR #478, opened 2018-12-14
Fix Rules apply function and rule to show selection

zurk:
Fixes https://github.com/src-d/style-analyzer/issues/461.

So there are fixes for two bugs. 
1. If no rule is applicable we by default predict label 0 and rule 0.
2. We pick up the wrong rule to show because of two lists inconsistency: list with rule indexes and list with fixes.

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -32,7 +32,6 @@ LineFix = NamedTuple("LineFix", (
     ("line_number", int),                     # line number for the comment
     ("suggested_code", str),                  # code line predicted by our model
     ("fixed_vnodes", List[VirtualNode]),      # VirtualNode-s with fixed y
-    ("winner_rules", List[int]),              # rule indices for fixed_vnodes
     ("confidence", int),                      # overall confidence in the prediction, 0-100
 ))
 
@@ -283,7 +282,6 @@ class FormatAnalyzer(Analyzer):
                 new_code_line=line_fix.suggested_code,  # code line suggested by our model
                 rules=rules.rules,                      # list of rules belonging to the model
                 fixed_vnodes=line_fix.fixed_vnodes,     # VirtualNode-s with changed y
-                winner_rules=line_fix.winner_rules,     # winner rule indices for fixed_vnodes
                 confidence=line_fix.confidence,         # overall confidence in the prediction
                 describe_rule=_describe_rule,           # function to format a rule as text
                 describe_change=_describe_change,       # function to format a change as text
@@ -297,6 +295,7 @@ class FormatAnalyzer(Analyzer):
         X, y, (vnodes_y, vnodes, vnode_parents, node_parents) = feature_extractor_output
         y_pred, rule_winners, new_rules, grouped_quote_predictions = rules.predict(
             X=X, vnodes_y=vnodes_y, vnodes=vnodes, feature_extractor=fe)
+        y_pred = rules.fill_missing_predictions(y_pred, y)
         if self.config["uast_break_check"]:
             y, y_pred, vnodes_y, rule_winners, safe_preds = filter_uast_breaking_preds(
                 y=y, y_pred=y_pred, vnodes_y=vnodes_y, vnodes=vnodes, files={file.path: file},
@@ -306,7 +305,7 @@ class FormatAnalyzer(Analyzer):
         assert len(y) == len(y_pred)
         assert len(y) == len(rule_winners)
         code_generator = CodeGenerator(fe, skip_errors=True)
-        new_vnodes = code_generator.apply_predicted_y(vnodes, vnodes_y, y_pred)
+        new_vnodes = code_generator.apply_predicted_y(vnodes, vnodes_y, rule_winners, new_rules)
         token_fixes = []
         for line_number, line in self._group_line_nodes(
                 y, y_pred, vnodes_y, new_vnodes, rule_winners):
@@ -321,7 +320,6 @@ class FormatAnalyzer(Analyzer):
                 line_number=line_number,        # line number for the comment
                 suggested_code=new_code_line,   # code line suggested by our model
                 fixed_vnodes=fixed_vnodes,      # VirtualNode-s with changed y
-                winner_rules=line_winners,      # winner rule indices for fixed_vnodes
                 confidence=confidence,          # overall confidence in the prediction, 0-100
             ))
         return token_fixes, new_vnodes
--- a/lookout/style/format/code_generator.py
+++ b/lookout/style/format/code_generator.py
@@ -2,12 +2,11 @@
 import logging
 from typing import List, Sequence
 
-import numpy
-
 from lookout.style.format.classes import (
     CLASS_INDEX, CLASS_REPRESENTATIONS, CLASSES, CLS_DOUBLE_QUOTE, CLS_NEWLINE, CLS_NOOP,
     CLS_SINGLE_QUOTE, CLS_SPACE_DEC, CLS_SPACE_INC, CLS_TAB_DEC, CLS_TAB_INC, CLS_TO_STR)
 from lookout.style.format.feature_extractor import FeatureExtractor
+from lookout.style.format.rules import Rules
 from lookout.style.format.virtual_node import VirtualNode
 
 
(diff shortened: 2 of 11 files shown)


lookout/style/format/analyzer.py
```diff
@@ -284,7 +282,7 @@ def _generate_comments_data(
             feature_extractor_output, bblfsh_stub: "bblfsh.aliases.ProtocolServiceStub",
             rules: Rules) -> Iterable[FixData]:
         X, y, (vnodes_y, vnodes, vnode_parents, node_parents) = feature_extractor_output
-        y_pred, rule_winners, new_rules = rules.predict(X=X, vnodes_y=vnodes_y, vnodes=vnodes,
+        y_pred, rule_winners, new_rules = rules.predict(X=X, y=y, vnodes_y=vnodes_y, vnodes=vnodes,
```
vmarkovtsev: Please rename it to `y_fallback`

lookout/style/format/code_generator.py
```diff
@@ -67,14 +71,19 @@ def apply_predicted_y(self, vnodes: Sequence[VirtualNode],
         :param vnodes_y: Sequence of the labeled `VirtualNode`-s corresponding to labeled samples.\
                          Should be ordered by start position value.
         :param y_pred: The model predictions for `vnodes_y` `VirtualNode`-s.
+        :param rule_winners: Sequence of applied rules to generate y_pred.
```
vmarkovtsev: If we have `rule_winners`, we can recover `y_pred` by looking at what each rule predicts.

lookout/style/format/rules.py
```diff
@@ -88,19 +88,19 @@ def apply(self, X_csr: csr_matrix, return_winner_indices=False,
...
-                 length as X containing (predictions, winner rule indices).
+                 length as X containing (predictions, winner rule indices). In case no rule was \
+                 triggered for feature row, corresponding result equals to -1.
         """
         X = X_csr.toarray()
         self._log.debug("predicting %d samples using %d rules", len(X), len(self._rules))
         rules = self._rules
         _compute_triggered = self._compute_triggered
-        prediction = numpy.zeros(len(X), dtype=numpy.int32)
+        prediction = -numpy.ones(len(X), dtype=numpy.int32)
```
vmarkovtsev:
```suggestion
        prediction = numpy.full(len(X), -1, dtype=numpy.int32)
```

lookout/style/format/rules.py
```diff
@@ -88,19 +88,19 @@ def apply(self, X_csr: csr_matrix, return_winner_indices=False,
...
         if return_winner_indices:
-            winner_indices = numpy.zeros(len(X), dtype=numpy.int32)
+            winner_indices = -numpy.ones(len(X), dtype=numpy.int32)
```
vmarkovtsev:
```suggestion
            winner_indices = numpy.full(len(X), -1, dtype=numpy.int32)
```

lookout/style/format/code_generator.py
```diff
@@ -67,14 +71,19 @@ def apply_predicted_y(self, vnodes: Sequence[VirtualNode],
...
         """
         result = []
-        for vnode, y_new in self._iterate_vnodes(vnodes, vnodes_y, y_pred):
+        for vnode, y_new, rule_winner in \
+                self._iterate_vnodes(vnodes, vnodes_y, y_pred, rule_winners):
             vnode = vnode.copy()
             if y_new != vnode.y:
                 vnode.y, vnode.y_old = y_new, vnode.y
+                if rule_winner:
+                    vnode.applied_rule = rules.rules[rule_winner]
```
vmarkovtsev: `else: vnode.applied_rule = None`?

vmarkovtsev:
`Rules.predict()` has diverged very far from the traditional `predict(X)`, also `applied_rule` bloats `VirtualNode`. This is OK for this time, but we really need to implement the annotations pattern which @m09 presented on Friday.

zurk:
Completely agree.

m09:
:+1:

vmarkovtsev:
@zurk Let's follow your first proposed solution with a separate PR and rebasing.

Regarding the second failure, I can see a dramatic increase in precision, niceee

lookout/style/format/analyzer.py
m09: as mentioned during our quick call this morning with @zurk, we should not use `y` during the predict phase since we are trying to remove the dependency on `y` in most of the codebase to be able to predict on unseen targets

lookout/style/format/code_generator.py
zurk: I use `hasattr()` to check if some rule were applied. If you have some arguments against I can change it.

lookout/style/format/code_generator.py
```diff
@@ -67,14 +68,19 @@ def apply_predicted_y(self, vnodes: Sequence[VirtualNode],
...
         :return: The list of VirtualNode-s with adjusted labels. "y_old" attribute of each node \
                  contains the previous label.
         """
         result = []
-        for vnode, y_new in self._iterate_vnodes(vnodes, vnodes_y, y_pred):
+        for vnode, y_new, applied_rule in \
+                self._iterate_vnodes(vnodes, vnodes_y, rule_winners, rules):
             vnode = vnode.copy()
             if y_new != vnode.y:
                 vnode.y, vnode.y_old = y_new, vnode.y
```
zurk: As Vadim asked I extract `y_new` from `rules.rules[rule_winners[i]].stats.cls` but I keep triplet `vnode.y, vnode.y_old, vnode.applied_rule`. However, it is possible to drop `vnode.y_old` and keep original target value in `vnode.y`. But it requires additional work. 
It is up to reviewers, I can do that. I do not feel a big need for this change because we want to add Hugo's annotation thing so all this stuff will be gone.

lookout/style/format/rules.py
```diff
@@ -207,6 +227,8 @@ def append_new_rule(labels: Tuple[int, ...], y_i: int, conf: float, support: int
         y_indices = {id(vnode): i for i, vnode in enumerate(vnodes_y)}
         for vnode1, vnode2, vnode3 in zip(vnodes, islice(vnodes, 1, None),
                                           islice(vnodes, 2, None)):
+            # What if we have prediction for one quote label and do not have for another?
+            # Should we change both or should we keep original?
```
zurk: It can happen because of this lines:
https://github.com/src-d/style-analyzer/pull/478/files#diff-fd9a5210475c52f533d210a1888cffa5R134

I think it is a question for Hugo. 
I will delete a comment when we resolve this situation.

lookout/style/format/visualizer/server.py
```diff
@@ -122,11 +124,13 @@ def return_features() -> Response:
     labeled_indices = {id(vnode): i for i, vnode in enumerate(vnodes_y)}
     app.logger.info("returning features of shape %d, %d" % X.shape)
     app.logger.info("length of rules: %d", len(rules))
+    # TODO (zurk): Add refuse_to_predict to visualization
```
zurk: I am not familiar with FrontEnd technologies, so I leave this todo for @m09 if you are so kind to help me with this. :) 
As soon as pure `y_pred` can have negative values I think this PR breaks visualization tool.

vmarkovtsev:
LGTM. @m09 I think the best will be to push the fix for the visual server here before merging.

lookout/style/format/rules.py
vmarkovtsev: Does still hold after we merged #482?

zurk: not sure. Let's wait for @m09 comment.

m09:
I'm on it

lookout/style/format/rules.py
m09: I think we should still pass it to harmonize quotes. I'll push with the visualizer fix

m09: Ok so after having observed the current result in visualizer I think this works fine (if we have prediction for only one quote it will undergo the uast breaking check anyway so if it's breaking it'll get discarded, and this way it can still fix whitespace, newlines, …)

zurk:
We are good to merge!

PR #480, opened 2018-12-17
Split FixData to LineFix and FileFix

zurk:
To solve the problem mentioned in https://github.com/src-d/style-analyzer/pull/478#issuecomment-447796005

TL;DR: now it is easy to process the file without comments. No need for an ad-hoc solution.

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -10,7 +10,7 @@ import bblfsh  # noqa: F401
 from jinja2 import Template
 from lookout.core.analyzer import Analyzer, ReferencePointer
 from lookout.core.api.service_analyzer_pb2 import Comment
-from lookout.core.api.service_data_pb2 import File
+from lookout.core.api.service_data_pb2 import Change, File
 from lookout.core.data_requests import DataService, \
     with_changed_uasts_and_contents, with_uasts_and_contents
 from lookout.core.lib import files_by_language, filter_files, find_deleted_lines, find_new_lines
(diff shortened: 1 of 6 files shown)


lookout/style/format/analyzer.py
```diff
@@ -28,18 +28,22 @@
...
-    ("all_vnodes", List[VirtualNode]),        # VirtualNode-s which correspond to this file
     ("fixed_vnodes", List[VirtualNode]),      # VirtualNode-s with fixed y
-    ("winner_rules", List[int]),              # Rule indices
+    ("winner_rules", List[int]),              # rule indices for fixed_vnodes
     ("confidence", int),                      # overall confidence in the prediction, 0-100
+))
+
+FileFix = NamedTuple("FileFix", (
+    ("error", str),                           # error message
+    ("line_fixes", Iterable[LineFix]),        # fixes for file lines. Can be empty.
```
vmarkovtsev: Sequence, not Iterable

lookout/style/format/analyzer.py
```diff
@@ -205,7 +213,8 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_service: D
         cls._log.info("trained %s", model)
         return model
 
-    def generate_fixes(self, data_service: DataService, changes: Sequence) -> Iterable[FixData]:
+    def generate_file_fixes(self, data_service: DataService, changes: Sequence,
```
vmarkovtsev: `changes: Sequence` of what? I see that we do not have the element type in the original code, can you please add it still?

lookout/style/format/analyzer.py
```diff
@@ -125,13 +129,17 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
-                confidence=fix_data.confidence))
+        for file_fix in self.generate_file_fixes(data_service, list(data["changes"])):
+            filename = file_fix.head_file.path
+            if file_fix.error:
+                comments.append(generate_comment(text=file_fix.error, line=0, confidence=100,
+                                                 filename=filename))
+
+            else:
+                comments.extend(generate_comment(
+                    text=self.render_comment_text(line_fix, file_fix), filename=filename,
```
vmarkovtsev: Can you please replace line_fix with an index inside file_fix? Otherwise, it is possible to pass a line fix which is not contained in the file fix.

lookout/style/format/analyzer.py
```diff
@@ -242,47 +251,48 @@ def generate_fixes(self, data_service: DataService, changes: Sequence) -> Iterab
...
+                    yield FileFix(error="", head_file=file, language=lang, feature_extractor=fe,
+                                  base_file=prev_file, file_vnodes=file_vnodes, line_fixes=fixes)
 
-    def render_comment_text(self, fix_data: FixData) -> str:
+    def render_comment_text(self, line_fix: LineFix, file_fix: FileFix) -> str:
         """
         Generate the text of the comment at the specified line.
 
-        :param fix_data: Full information required to render a comment text.
+        :param line_fix: Information about line fixes required to render a comment text.
```
vmarkovtsev: Hence the index instead of the object here

lookout/style/format/tests/test_quality_report.py
```diff
@@ -144,7 +146,7 @@ def test_eval_aggregate(self):
         output = "\n".join(output)
         output = output[:output.find("# Model report for https://github.com/jquery/jquery")]
         metrics = _get_precision_recall_f1_support(output)
-        self.assertEqual(
+        self.assertAlmostEqual(
```
vmarkovtsev: The precision allows the check to be exact. If we float in distant signs, there is an error in our code because the process must be fully deterministic.

zurk: It was `0.8397747184701835` vs `0.8397747184701836`

zurk: And not for all python versions.

zurk:
@vmarkovtsev PTAL. I set `delta=1e-15` for `assertAlmostEqual` because it looks like some random difference related to rounding or something similar.

PR #482, opened 2018-12-17
Fix quote UAST check

m09:
This adds UAST check to quote predictions. Goal is to be able to detect stuff like `a = "'b'"` being changed to `a = ''b''` and breaking everything.

### DIFF
--- a/lookout/style/format/feature_extractor.py
+++ b/lookout/style/format/feature_extractor.py
@@ -15,7 +15,8 @@ from sklearn.feature_selection import SelectKBest, VarianceThreshold
 
 from lookout.style.format.classes import (
     CLASS_INDEX, CLASS_PRINTABLES, CLASS_REPRESENTATIONS, CLS_DOUBLE_QUOTE, CLS_NEWLINE, CLS_NOOP,
-    CLS_SINGLE_QUOTE, CLS_SPACE, CLS_SPACE_DEC, CLS_SPACE_INC, CLS_TAB, CLS_TAB_DEC, CLS_TAB_INC)
+    CLS_SINGLE_QUOTE, CLS_SPACE, CLS_SPACE_DEC, CLS_SPACE_INC, CLS_TAB, CLS_TAB_DEC, CLS_TAB_INC,
+    INDEX_CLS_TO_STR)
 from lookout.style.format.features import (  # noqa: F401
     Feature, FEATURE_CLASSES, FeatureGroup, FeatureId, FeatureLayout, Layout,
     MultipleValuesFeature, MutableFeatureLayout, MutableLayout)
@@ -359,6 +360,11 @@ class FeatureExtractor:
         """
         return ["".join(CLASS_PRINTABLES[label] for label in labels)
                 for labels in self.labels_to_class_sequences]
+
+    def label_to_str(self, label: int) -> str:
+        """Convert a label to string."""
+        return "".join(INDEX_CLS_TO_STR[cls]
+                       for cls in self.labels_to_class_sequences[label])
 
     def _create_neighbours(self, vnodes: Sequence[VirtualNode], vnodes_y: Sequence[VirtualNode],
                            parents: Mapping[int, bblfsh.Node],
--- a/lookout/style/format/postprocess.py
+++ b/lookout/style/format/postprocess.py
@@ -1,5 +1,5 @@
 """Postprocess predictions of the model."""
-from logging import Logger
+from logging import getLogger
 from typing import Dict, Mapping, MutableMapping, Optional, Sequence, Tuple  # noqa: F401
 
 import bblfsh
@@ -7,9 +7,12 @@ from lookout.core.api.service_data_pb2 import File
 from lookout.core.data_requests import parse_uast
 import numpy
 
-from lookout.style.format.classes import CLS_DOUBLE_QUOTE, CLS_SINGLE_QUOTE, INDEX_CLS_TO_STR
 from lookout.style.format.feature_extractor import FeatureExtractor
+from lookout.style.format.rules import QuotedNodeTripleMapping
 from lookout.style.format.virtual_node import VirtualNode
+
+
+_log = getLogger("Postprocess")
 
 
 def check_uasts_are_equal(uast1: bblfsh.Node, uast2: bblfsh.Node) -> bool:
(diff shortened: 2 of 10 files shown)


zurk:
I thought it can be checked with regular `uast_break_check`, no?

m09:
Regular `filter_uast_breaking_preds` only tests one change at a time. Quotes should be tested in pairs.

lookout/style/format/rules.py
```diff
@@ -165,10 +171,30 @@ def _get_composite(feature_extractor: FeatureExtractor, labels: Tuple[int, ...])
...
+        for vnode1, vnode2, vnode3 in zip(vnodes, islice(vnodes, 1, None),
+                                          islice(vnodes, 2, None)):
+            if (id(vnode1) not in y_indices or id(vnode3) not in y_indices or vnode2.node is None
+                    or vnode1.y[-1] not in quotes_classes or vnode3.y[0] != vnode1.y[-1]):
+                continue
+            vnode2_roles = frozenset(role_name(role_id) for role_id in vnode2.node.roles)
+            if "STRING" in vnode2_roles:
+                grouped_predictions[id(vnode1)] = vnode1, vnode2, vnode3
+                grouped_predictions[id(vnode3)] = None
+        return grouped_predictions
```
zurk: Maybe it is better to return one index for `vnodes` sequence instead of 3 elements from it? The index of `vnode2`. You can be always sure that there are quotes tokes on the right and left.

lookout/style/format/postprocess.py
```diff
@@ -76,18 +78,25 @@ def _parse_code(parent: bblfsh.Node, content: str, stub: "bblfsh.aliases.Protoco
...
     else:
         result = None
-        logger.warning("skipped file %s, due to errors in parsing the whole content", path)
+        _log.warning("skipped file %s, due to errors in parsing the whole content", path)
     for descendant in descendants:
         parsing_cache[id(descendant)] = result
     return result
 
 
+def _prediction_to_text(prediction: int, feature_extractor: FeatureExtractor) -> str:
```
vmarkovtsev: Should it go to descriptions.py?

lookout/style/format/postprocess.py
```diff
@@ -145,21 +170,21 @@ def filter_uast_breaking_preds(
             # to handle mixed indentations, we include the `VirtualNode` following the predicted
             # one in the output predicted string, and start the rest of the sequence one
             # `VirtualNode` further to avoid its repetitions
-            content_after = content_before[:vn_y.start.offset] \
-                + output_pred \
-                + content_before[vn_y.start.offset + len(vn_y.value)
-                                 + len(vnodes[cur_i + 1].value):]
+            content_after = (file_content[:vn_y.start.offset]
+                             + output_pred
+                             + file_content[vn_y.start.offset + len(vn_y.value)
```
vmarkovtsev: Let's move `vn_y.start.offset + len(vn_y.value) + len(vnodes[cur_i + 1].value)` to a new variable and use it here.

lookout/style/format/postprocess.py
```diff
@@ -104,35 +113,51 @@ def filter_uast_breaking_preds(
...
+        # quote types are special cased
+        if id(vn_y) in grouped_quote_predictions:
+            pred_string = _prediction_to_text(pred, feature_extractor)
+            group = grouped_quote_predictions[id(vn_y)]
+            # already handled with the previous vnode
+            if group is None:
+                continue
+            vnode1, vnode2, vnode3 = group
+            content1, content2, content3 = map(lambda v: file_content[v.start.offset:v.end.offset],
+                                               (vnode1, vnode2, vnode3))
```
zurk: Why not `content1, content2, content3 =  vnode1.value, vnode2.value, vnode3.value`? should be the same.

lookout/style/format/rules.py
```diff
@@ -116,9 +116,11 @@ def apply(self, X_csr: csr_matrix, return_winner_indices=False,
...
         return prediction
 
-    def predict(self, X: numpy.ndarray, vnodes_y: Sequence[VirtualNode],
-                vnodes: Sequence[VirtualNode], feature_extractor: FeatureExtractor,
-                ) -> Tuple[numpy.ndarray, numpy.ndarray, "Rules"]:
+    def predict(
+            self, X: numpy.ndarray, vnodes_y: Sequence[VirtualNode], vnodes: Sequence[VirtualNode],
+            feature_extractor: FeatureExtractor,
+            ) -> Tuple[numpy.ndarray, numpy.ndarray, "Rules",
+                       Mapping[int, Optional[Tuple[VirtualNode, VirtualNode, VirtualNode]]]]:
```
vmarkovtsev: Let's define a new type:

```python
QuotedNodeTriple = NamedTuple("QuotedNodeTriple", [["left", VirtualNode], ["target", VirtualNode], ["right", VirtualNode]])
```

lookout/style/format/rules.py
m09: This is basically an iterator. Changing it to only return the index is akin to only implementing half of the iterator's logic for me. I'd rather return a triple to implement the full iterator logic

lookout/style/format/postprocess.py
m09: I agree that it's not readable. I'm currently discussing this part of the function with Warren to see how I can further refactor it.

vmarkovtsev:
@zurk all good now?

lookout/style/format/feature_extractor.py
```diff
@@ -360,6 +361,11 @@ def composite_class_printables(self) -> List[str]:
         return ["".join(CLASS_PRINTABLES[label] for label in labels)
                 for labels in self.labels_to_class_sequences]
 
+    def label_to_str(self, label: int) -> str:
+        """Convert a label to string."""
+        return "".join(INDEX_CLS_TO_STR[cls]
+                       for cls in self.labels_to_class_sequences[label])
```
zurk: Mmm, I did not mention it earlier, but we have this guy in `FeatureExtractor`: 
https://github.com/src-d/style-analyzer/blob/a38de6d64803ae85ed835411d327fb27e954e123/lookout/style/format/feature_extractor.py#L344

zurk: ok, it is different.

PR #483, opened 2018-12-18
Fix quote prediction & tests

m09:
Slipped through yesterday


m09:
I merged w/o review since it's a simple bugfix.

PR #485, opened 2018-12-18
Fix empty string labeling

### LINKED ISSUE #481: Empty string creates one sample less than non-empty strings, and more labels
Since the sequence labels and the label merging, `'b' + 3;` and `'' + 3;` have a different number of samples:
In the first case:
- `'`
- `'` and `space`
- `space`
- `noop`
In the second:
- `'`, `'` and `space`
- `space`
- `noop`

which basically means that empty strings are not handled like the others and create many labels depending on their surroundings.

m09:
Fixes #481.

### DIFF
--- a/lookout/style/format/tests/test_quality_report.py
+++ b/lookout/style/format/tests/test_quality_report.py
@@ -146,7 +146,7 @@ class QualityReportTests(PretrainedModelTests):
         output = "\n".join(output)
         output = output[:output.find("# Model report for https://github.com/jquery/jquery")]
         metrics = _get_precision_recall_f1_support(output)
-        expected_metrics = (0.9191398589201619, 0.6745843230403801, 0.7385002866178716, 2947)
+        expected_metrics = (0.9220615191829985, 0.673337856173677, 0.7376592242904151, 2948)
         self.assertEqual(len(metrics), len(expected_metrics))
         for metric_value, expected_metric_value in zip(metrics, expected_metrics):
             self.assertAlmostEqual(metric_value, expected_metric_value, delta=1e-15)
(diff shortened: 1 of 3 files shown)


lookout/style/format/virtual_node.py
```diff
@@ -117,6 +117,14 @@ def from_node(node: bblfsh.Node, file: str, path: str,
         """
         outer_token = file[node.start_position.offset:node.end_position.offset]
         if not node.token:
+            if outer_token == "''" or outer_token == '""':
+                start = Position.from_bblfsh_position(node.start_position)
+                middle = Position(offset=start.offset, line=start.line, col=start.col + 1)
```
zurk:
```suggestion
                middle = Position(offset=start.offset + 1, line=start.line, col=start.col + 1)
```

lookout/style/format/tests/test_quality_report.py
```diff
@@ -146,7 +146,7 @@ def test_eval_aggregate(self):
         output = "\n".join(output)
         output = output[:output.find("# Model report for https://github.com/jquery/jquery")]
         metrics = _get_precision_recall_f1_support(output)
-        expected_metrics = (0.9191398589201619, 0.6745843230403801, 0.7385002866178716, 2947)
+        expected_metrics = (0.9220615191829985, 0.673337856173677, 0.7376592242904151, 2948)
```
vmarkovtsev: Precision +, Recall - as expected, good

lookout/style/format/virtual_node.py
```diff
@@ -117,6 +117,14 @@ def from_node(node: bblfsh.Node, file: str, path: str,
         """
         outer_token = file[node.start_position.offset:node.end_position.offset]
         if not node.token:
+            if outer_token == "''" or outer_token == '""':
```
vmarkovtsev: We are going to have fun time porting to other languages which distinguish between single and double quotes...

m09: We'll also have a lot of fun when we'll port to languages with `"""` and other quoting sequences because bblfsh doesn't help us at all in this case to know where the delimiters are so we'll have to go full-heuristic.

Issue #502, opened 2018-12-20
Remove files_filtering.py and use lookout-sdk-ml


vmarkovtsev:
`style-analyzer/lookout/style/format/files_filtering.py`duplicates https://github.com/src-d/lookout-sdk-ml/blob/master/lookout/core/lib.py#L101 as @warenlg found. We need to add the missing regexp filtering feature and switch to using the SDK function.

zurk:
Is it solved, guys? There is no such file `style-analyzer/lookout/style/format/files_filtering.py` anymore.

vmarkovtsev:
Not really, @warenlg is pending to rebase what's left in his older commits.

warenlg:
Yes, I'm on it right now

warenlg:
Order of review required:
1. https://github.com/src-d/lookout-sdk-ml/pull/64
2. lookout release
3. https://github.com/src-d/style-analyzer/pull/652

PR #508, opened 2018-12-26
Remove files_filtering and use lookout-sdk-ml

warenlg:
Fixes https://github.com/src-d/style-analyzer/issues/502

This PR should be reviewed/merged once https://github.com/src-d/lookout-sdk-ml/pull/31 is merged and the new lookout-sdk released. Before, the tests won't pass anyway.


lookout/style/format/analyzer.py
```diff
@@ -168,7 +168,6 @@ def train(cls, ptr: ReferencePointer, config: Mapping[str, Any], data_service: D
             except KeyError:
                 _log.warning("language %s is not supported, skipped", language)
                 continue
-            files = filter_files(files, lang_config["line_length_limit"], _log)
```
vmarkovtsev: Why don't we filter here?

PR #512, opened 2019-01-08
Fix report

EgorBu:
* report for all samples
* handle empty report
* minor improvements to avoid hardcoding

### DIFF
--- a/lookout/style/format/benchmarks/top_repos_quality.py
+++ b/lookout/style/format/benchmarks/top_repos_quality.py
@@ -1,5 +1,6 @@
 """Measure quality on several top repositories."""
 from argparse import ArgumentParser, Namespace
+from collections import OrderedDict
 from datetime import datetime
 import functools
 import importlib
@@ -49,6 +50,8 @@ REPOSITORIES = [
     "https://github.com/atom/atom 108b23210759a8c5b2f51ac99659be5dc31a7371 a3c320dd707b915da2192427bcceea166edbd6d4",  # noqa: E501
     "https://github.com/nodejs/node 6eda924c189e44a36fc97a7cfae41b69483d5bfb 315b1c656cee39c989015cc2b17fe8c864dbc3dd",  # noqa: E501
 ]
+
+FLOAT_PRECISION = ".3f"
 
 
 class AnalyzerContextManager:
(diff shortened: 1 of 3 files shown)


lookout/style/format/benchmarks/top_repos_quality.py
```diff
@@ -231,6 +233,7 @@ def calc_avg(arr: Sequence[Sequence], col: int) -> float:
     ("recall", float),
     ("full_recall", float),
     ("f1", float),
+    ("full_f1", float),
```
zurk: I think it is worth to make a comment about `full_` prefix because It requires some time for me to remember what it means. 
I suggest to add something like:
```python
# Metrics for the quality report. Metrics are calculated on the samples subset 
# where predictions were made.  `full_` prefix means that metric was calculated 
# on all available samples. `ppcr` means predicted positive condition rate and
# shows the ratio of samples where the model was able to predict. 
```

before `NamedTuple` defenition.

lookout/style/format/benchmarks/top_repos_quality.py
```diff
@@ -159,7 +161,7 @@ def __init__(self, quality: Optional[str] = None, model: Optional[str] = None) -
 def measure_quality(repository: str, from_commit: str, to_commit: str, port: int,
                     review_config: dict, train_config: dict) -> QualityReport:
     """
-    Generate quality and model reports for a repository.
+    Generate quality and model reports for a repository. If it fails - it returns empty reports.
```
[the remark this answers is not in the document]
EgorBu: it returns `:return: Reports.`

lookout/style/format/benchmarks/top_repos_quality.py
```diff
@@ -231,10 +234,18 @@ def calc_avg(arr: Sequence[Sequence], col: int) -> float:
...
     ("support", int),
     ("full_support", int),
 ))
+Metrics.__doc__ = """Metrics for the quality report. Metrics are calculated on the samples subset
+where predictions were made.  `full_` prefix means that metric was calculated on all available
+samples. Without `full_` means that metric was calculated only on samples where it has prediction
+from the model. `ppcr` means predicted positive condition rate and shows the ratio of samples where
+the model was able to predict.
+"""
+#
```
vmarkovtsev: Forgot to remove `#`

Issue #511, opened 2019-01-08
Fix the last-minute problems with the reports


vmarkovtsev:
- [x] Fix recall and f1 being the same as precision
- [x] Check for None in the summary report and handle gracefully
- [x] Investigate and fix (if easy to fix) the crash on meteor

vmarkovtsev:
\+ https://github.com/src-d/ml-backlog/issues/1

vmarkovtsev:
Reports wer efixed in https://github.com/src-d/style-analyzer/pull/512
@EgorBu has regenerated the summary for the quality report.

The problem with meteor is some weird Unicode, is fixed in https://github.com/src-d/style-analyzer/pull/519

zurk:
Ok, with fix https://github.com/src-d/style-analyzer/pull/527 meteor should work fine and I am going to check it today.

zurk:
the meteor is still running (attributes prunning)
I am going to run another report collection without attributes prunning.

zurk:
Fresh raw report summaries without attributes pruning for https://github.com/src-d/style-analyzer/commit/171ebabf5399d7fe2ab812903a58e854a1b1e519

# How to reproduce: 
1. Disable pruning in `analyzer.py`: `"prune_attributes": False, "uncertain_attributes": False`
2. Run `make report-quality`

# Train report

| repo               |   precision |   recall |   full_recall |    f1 |   full_f1 |   ppcr |   support |   full_support |   Rules Number |   Average Rule Len |
|:-------------------|------------:|---------:|--------------:|------:|----------:|-------:|----------:|---------------:|---------------:|-------------------:|
| 30-seconds-of-code |       0.938 |    0.938 |         0.905 | 0.938 |     0.921 |  0.964 |     65310 |          67737 |        151     |            7.05298 |
| axios              |       0.960 |    0.960 |         0.959 | 0.960 |     0.959 |  1.000 |     21126 |          21130 |        143     |            7.27273 |
| carlo              |       0.961 |    0.961 |         0.909 | 0.961 |     0.934 |  0.945 |      5227 |           5529 |         78     |            6.20513 |
| citgm              |       0.978 |    0.978 |         0.347 | 0.978 |     0.513 |  0.355 |      7797 |          21941 |         14     |            6       |
| create-react-app   |       0.947 |    0.947 |         0.888 | 0.947 |     0.916 |  0.938 |     15687 |          16718 |        101     |            6.55446 |
[...]

PR #519, opened 2019-01-09
Fix line positions bug and add test

zurk:
Related to last point of https://github.com/src-d/style-analyzer/issues/511

The main problem that for some files with specific unicode symbols like `\u2028` and `\u2029` we have bad line positions.

Should be fixed in `FeatureExtractor._parse_file()`


zurk:
I was able to fix the bug.
Moral: always use `.splitlines()` function insdead of `.split("\n")`

@vmarkovtsev or @m09 please review.

PR #527, opened 2019-01-11
Add test for virtual nodes position in the end of file

zurk:
New bug, discovered after https://github.com/src-d/style-analyzer/pull/519

### DIFF
--- a/lookout/style/format/feature_extractor.py
+++ b/lookout/style/format/feature_extractor.py
@@ -668,13 +668,21 @@ class FeatureExtractor:
         :return: list of `VirtualNode`-s and the parents.
         """
         # build the line mapping
-        lines = contents.splitlines()
+        lines = contents.splitlines(keepends=True)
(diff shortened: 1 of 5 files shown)


zurk:
So the problem is that all file contains `\r\n` as line separator and the last line ends with just '\n'

zurk:
It looks like I was able to fix the issue. 

Some points:
1. There is a lot of unobvious changes in code so I add comments. If you do not understand them, we should change them it. I hope they are fine so I do not comment changes here more. 
2. Fix is also not really beautiful, but I do not want to change a lot in our core function `_parse_file()`. So what I did is restoring the same value as we had in `line_offsets` variable before changes. 
2. I found one more place where we still use `.split("\n")` instead of `.splitlines()` and it is bad. It is unrelated to the initial bug, so I fix it in another PR. 

@vmarkovtsev or @m09 ready for review.

PR #530, opened 2019-01-14
Replace .split("\n") with .splitlines() in_classify_vnodes()

zurk:
This is the last part of `split -> splitlines` refactoring. 

This change nothing for a regular code but fixes processing for any weird Unicode newlines that is used in JS.

### DIFF
--- a/lookout/style/format/feature_extractor.py
+++ b/lookout/style/format/feature_extractor.py
@@ -461,12 +461,10 @@ class FeatureExtractor:
                     node.y = (CLASS_INDEX[CLS_DOUBLE_QUOTE],)
                 yield node
                 continue
-            lines = node.value.split("\r\n")
-            if len(lines) > 1:
-                sep = "\r\n"
-            else:
-                lines = node.value.split("\n")
-                sep = "\n"
+            lines = node.value.splitlines(keepends=True)
+            if lines[-1].splitlines()[0] != lines[-1]:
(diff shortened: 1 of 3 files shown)


vmarkovtsev:
@zurk CI failed

zurk:
yeah, in the end, It requires the same tricks as for https://github.com/src-d/style-analyzer/pull/527

lookout/style/format/feature_extractor.py
```diff
@@ -484,17 +482,16 @@ def _classify_vnodes(self, nodes: Iterable[VirtualNode], path: str) -> Iterable[
...
-                newline = line + sep
                 start_offset = node.start.offset + line_offset
                 start_col = node.start.col if i == 0 else 1
                 lineno = node.start.line + i
                 yield VirtualNode(
-                    newline,
+                    line,
                     Position(start_offset, lineno, start_col),
-                    Position(start_offset + len(newline), lineno, start_col + len(newline)),
+                    Position(start_offset + len(line), lineno + 1, 1),
```
zurk: @vmarkovtsev and @m09 please pay attention here.
I think it was a bug because it behaved differently in comparison with pure `_parse_file()` call: token with a new line ends at the beginning of next line. 

If it was on purpose we should change it in `_parse_file()` method. 

I see that this change can affect Hugo's visualizer, some of the features like `_FeatureDiffLine`.

Issue #526, opened 2019-01-10
Review all TODOs and FIXMEs


vmarkovtsev:
Convert to issues as needed.

EgorBu:
List of `TODO`s:
```python
--
./lookout/style/format/analyzer.py-                "\n\t".join("%-55s %.5E" % (fe.feature_names[i], importances[i])
./lookout/style/format/analyzer.py-                            for i in numpy.argsort(-importances)[:25] if importances[i] > 1e-5))
./lookout/style/format/analyzer.py-            submit_event("%s.train.%s.rules" % (cls.name, language), len(trainable_rules.rules))
./lookout/style/format/analyzer.py:            # TODO(vmarkovtsev): save the achieved precision, recall, etc. to the model
./lookout/style/format/analyzer.py-            # throw away imprecise classes
./lookout/style/format/analyzer.py-            if trainable_rules.rules.rules:
./lookout/style/format/analyzer.py-                model[language] = trainable_rules.rules
Binary file ./lookout/style/format/github_integration/lookout-dummy-analyzer_linux_amd64/dummy matches
Binary file ./lookout/style/format/github_integration/lookout_linux_amd64/lookoutd matches
--
./lookout/style/format/tests/top_repos_quality.py-# from lookout.style.format.tests.test_analyzer_integration import TestAnalyzer
./lookout/style/format/tests/top_repos_quality.py-
./lookout/style/format/tests/top_repos_quality.py-
./lookout/style/format/tests/top_repos_quality.py:# TODO: add after https://github.com/src-d/style-analyzer/issues/329 fixed
./lookout/style/format/tests/top_repos_quality.py-"""https://github.com/meteor/meteor c3309b123a7220ac24cbe73661184ee946bca01f 62fa9927ce34cff064cc3991439553e7c52b5258
./lookout/style/format/tests/top_repos_quality.py-https://github.com/webpack/webpack babe736cfa1ef7e8014ed32ba4a4ec38049dce14 3e74cb428af04eedac60ae13d2420d2b5bd3bde1"""  # noqa: E501
./lookout/style/format/tests/top_repos_quality.py:# TODO: add after https://github.com/bblfsh/client-python/issues/89 fixed
./lookout/style/format/tests/top_repos_quality.py-"""https://github.com/vuejs/vue b7105ae8c9093e36ec89a470caa3b78bda3ef467 db1d0474997e9e60b8b0d39a3b7c2af55cfd0d4a
[...]

PR #535, opened 2019-01-16
Complete several TODOs

EgorBu:
* handle custom modification functions as well as regexp 
https://github.com/src-d/style-analyzer/issues/526#issuecomment-454801024
* use dulwich in the future
https://github.com/src-d/style-analyzer/issues/526#issuecomment-454804489
* switch to numpy arrays
https://github.com/src-d/style-analyzer/issues/526#issuecomment-454808654
https://github.com/src-d/style-analyzer/issues/526#issuecomment-454809317

### DIFF
--- a/lookout/style/format/benchmarks/generate_smoke.py
+++ b/lookout/style/format/benchmarks/generate_smoke.py
@@ -103,7 +103,6 @@ def generate_smoke_entry(inputpath: str, outputpath: str, force: bool = False) -
         writer = csv.DictWriter(index_file, fieldnames=["repo", "style", "from", "to"])
         writer.writeheader()
         for style_name, (pattern, repl) in js_format_rules.items():
-            # TODO(zurk): handle custom modification functions as well as regexp.
             pattern = re.compile(pattern)
             for repopath in repopaths:
                 repo = Repo(str(repopath))
--- a/lookout/style/format/benchmarks/top_repos_quality.py
+++ b/lookout/style/format/benchmarks/top_repos_quality.py
@@ -9,10 +9,10 @@ import json
 import logging
 import logging.handlers
 import os
-import subprocess
 import tempfile
 from typing import Iterable, NamedTuple, Optional, Sequence, Type, Union
 
+from dulwich import porcelain
 from lookout.core import slogging
 from lookout.core.analyzer import Analyzer
 from lookout.core.cmdline import ArgumentDefaultsHelpFormatterNoNone, create_model_repo_from_args
@@ -20,6 +20,7 @@ from lookout.core.data_requests import DataService
 from lookout.core.event_listener import EventListener
 from lookout.core.manager import AnalyzerManager
 from lookout.core.test_helpers import server
+import numpy
 from tabulate import tabulate
 
 from lookout.style.format.benchmarks.general_report import QualityReportAnalyzer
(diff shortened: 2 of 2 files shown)


lookout/style/format/benchmarks/generate_smoke.py
```diff
@@ -103,14 +104,20 @@ def generate_smoke_entry(inputpath: str, outputpath: str, force: bool = False) -
         writer = csv.DictWriter(index_file, fieldnames=["repo", "style", "from", "to"])
         writer.writeheader()
         for style_name, (pattern, repl) in js_format_rules.items():
-            # TODO(zurk): handle custom modification functions as well as regexp.
-            pattern = re.compile(pattern)
+            if isinstance(pattern, str):
```
zurk: I am not sure about its signature (input parameters mostly). For now, I think that `repl ` argument is redundant and it should look like 
```
new_style_code = modify_style(initial_code)
```

I think we can keep this TODO in code as soon as we do not have any function for style modification.
When (and If) we actually need style modification functions, we can resolve it with a better understanding of what we actually need.

If we want to close all todos in the code we can just delete it. I think it is a minor one.

lookout/style/format/benchmarks/top_repos_quality.py
```diff
@@ -136,10 +137,8 @@ def ensure_repo(repository: str, storage_dir: str) -> str:
     if os.path.exists(repository):
         return repository
     # clone repository
-    # TODO: use dulwich in the future
     git_dir = os.path.join(storage_dir, get_repo_name(repository))  # location for code
-    cmd = "git clone --single-branch --branch master %s %s" % (repository, git_dir)
-    subprocess.check_call(cmd.split())
+    porcelain.clone(repository, git_dir)
```
zurk: it is not possible to clone only `--single-branch` via Dulwich, isn't it?

lookout/style/format/benchmarks/top_repos_quality.py
```diff
@@ -211,26 +210,18 @@ def wrapped_capture_quality_report(*args, **kwargs):
 
 def calc_weighted_avg(arr: Sequence[Sequence], col: int, weight_col: int = 5) -> float:
     """Calculate average value in `col` weighted by column `weight_col`."""
-    numerator, denominator = 0, 0
-    # TODO: switch to numpy arrays
-    for row in arr:
-        numerator += row[col] * row[weight_col]
-        denominator += row[weight_col]
+    arr = numpy.array(arr)
```
zurk: `arr` here can contain ints, floats as well as strings. Does this code work well in this case?

lookout/style/format/benchmarks/generate_smoke.py
```diff
@@ -103,14 +104,20 @@ def generate_smoke_entry(inputpath: str, outputpath: str, force: bool = False) -
...
+                pattern_ = re.compile(pattern)
+
+                def pattern(repl, code):
+                    return re.sub(pattern_, repl, code)
+            else:
+                assert isinstance(pattern, FunctionType), "Expected function, got %s" % \
```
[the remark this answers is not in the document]
EgorBu: https://github.com/src-d/style-analyzer/pull/535#discussion_r248968605

lookout/style/format/benchmarks/top_repos_quality.py
EgorBu: `.astype(float)` added to be sure that it works.

zurk: ok, I hope it is fine now.

PR #537, opened 2019-01-17
Test the real package

### LINKED ISSUE #532: pip install . is not tested
Currently, our CI steps and docker images all use `pip install -e .`. It leads to bugs such as #528.

vmarkovtsev:
Fixes #532


vmarkovtsev:
@m09 I reached the point here that we check the real package, and hit the absence of many-many files. I will continue tomorrow.

PR #541, opened 2019-01-21
Limit the maximum size of the training and test sets

vmarkovtsev:


lookout/style/format/analyzer.py
```diff
@@ -101,6 +101,7 @@ class FormatAnalyzer(Analyzer):
             "test_dataset_ratio": 0.0,
             "line_length_limit": 500,
             "lower_bound_instances": 500,
+            "overall_size_limit": 2 << 20,  # 2 MB
```
zurk: why do you set exactly this number?

vmarkovtsev: Taken from jQuery. I guess everything which is bigger starts to cause trouble.

vmarkovtsev:
@m09 Can you please relaunch the min_samples_leaf experiments with this PR merged - it is going to change the numbers.

m09:
sure!

PR #545, opened 2019-01-22
Rename eval-smoke-dataset parser variable and fix bblfsh arg

zurk:
The --bblfsh should not contain default value for smoke dataset evaluation because it blocks LOOKOUT_BBLFSHD environment variable usage. 
Without this change it is harder to use from docker container.


zurk:
@vmarkovtsev can be merged.

PR #549, opened 2019-01-24
Remove double feature selection

### LINKED ISSUE #548: [question] Double feature selection
I have spotted

```python
X_test, selected_features = fe.select_features(X_test, y_test)
```

in the train function. I feel that we should not select features for the second time. Is it correct?

m09:
Fixes #548.


vmarkovtsev:
@zurk FYI

zurk:
So... if we run
```python
X_train, y_train, _ = fe.extract_features(train_files)
X_train, selected_features = fe.select_features(X_train, y_train)
```
next run of 
```python
X_test, y_test, _ = fe.extract_features(test_files)
```
returns only selected features?

m09:
yes

PR #555, opened 2019-01-28
Reduce the number of hyperopt iterations to 50

vmarkovtsev:


zurk:
Why?

vmarkovtsev:
Because all the repos I saw, from small to big, always converge before 50

Issue #558, opened 2019-01-29
Rules include redundant attributes


vmarkovtsev:
I have spotted the same pattern several times. Our rules may have several attributes which are described as

```
-1.reserved = '(' and not in {';'}
```

or similar. This includes our paper which I edited by hand. The redundant part "not in" should be removed.

PR #560, opened 2019-01-29
Categorical pruning

### LINKED ISSUE #559: Simplify rule descriptions implementation
I digged the current implementation in `descriptions.py` and found it overcomplicated and very hard to understand. We should get rid of `@singledispatch` because it goes against "explicit is better than implicit" Zen of Python. The logic should be organized using the common Python machinery: dicts, functions and regular decorators. No metaclasses, no generics, no single dispatch.

vmarkovtsev:
Closes #559 and #558

### DIFF
--- a/lookout/style/format/descriptions.py
+++ b/lookout/style/format/descriptions.py
@@ -207,7 +207,7 @@ def _describe_rule_parts_ordinal(_: OrdinalFeature, name: str, splits: List[Rule
     """
     Describe a part of an ordinal rule in natural language.
 
-    :param feature: The feature used for the splits to describe.
+    :param _: The feature used for the splits to describe.
     :param name: The name to use for the feature used in the split.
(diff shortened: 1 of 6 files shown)


lookout/style/format/descriptions.py
```diff
@@ -216,6 +225,32 @@ def _describe_rule_parts_ordinal(feature: OrdinalFeature, name: str,
     return "%s ≤ %d" % (name, floor(threshold))
 
 
+RULE_SPLITS_DESCRIBERS = {
+    BagFeature: _describe_rule_splits_bag,
+    CategoricalFeature: _describe_rule_parts_categorical,
+    OrdinalFeature: _describe_rule_parts_ordinal,
```
zurk: different name patterns look suspicious.
let's rename to make them uniform?

PR #567, opened 2019-01-30
Train format analyzer on 20% different code

vmarkovtsev:

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -6,7 +6,7 @@ import logging
 import os
 from pprint import pformat
 import random
-from typing import Any, Iterable, List, Mapping, NamedTuple, Sequence, Tuple
+from typing import Any, Iterator, List, Mapping, NamedTuple, Sequence, Tuple
 import warnings
 
 import bblfsh  # noqa: F401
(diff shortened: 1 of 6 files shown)


vmarkovtsev:
@m09 @zurk I need to merge this asap so please post-review when you have time and I will fix your suggestions.

lookout/style/format/analyzer.py
```diff
@@ -163,25 +165,73 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+            for file in filter_files(head_files, lang_config["line_length_limit"],
+                                     lang_config["overall_size_limit"], log=_log):
+                head_lines = len(file.content.splitlines())
+                overall_lines += head_lines
+                try:
+                    prev_file = base_files_by_lang[language][file.path]
+                except KeyError:
+                    changed_lines += head_lines
+                else:
+                    changed_lines += len(find_new_lines(prev_file, file))
```
m09: we should probably check for deletions also (with `find_deleted_lines`) since a 20% change in this direction (less data) would probably also require a retrain.

lookout/style/format/analyzer.py
```diff
@@ -163,25 +165,73 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+            ratio = changed_lines / (overall_lines or 1)
+            _log.debug("check %s ratio: %.3f", language, ratio)
+            if ratio > lang_config["lines_ratio_train_trigger"]:
```
m09: There might be a bug here: since we put `0.8` in `lang_config["lines_ratio_train_trigger"]` and not `0.2`, we should use `if ratio > 1 - lang_config["lines_ratio_train_trigger"]`

lookout/style/format/analyzer.py
```diff
@@ -163,25 +165,73 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+                _log.info("%s triggers the training (%.3f)", language, ratio)
```
zurk: `_log.info("%s triggers the training with changes ratio %.3f", language, ratio)`

Issue #522, opened 2019-01-10
Remove global indentation algorithm from CodeGenerator


vmarkovtsev:
Local indentation is clearly better so we drop the global one and simplify the code.

zurk:
The first part in PR: https://github.com/src-d/style-analyzer/pull/584

To continue with this issue I should implement https://github.com/src-d/style-analyzer/issues/523 first

zurk:
After a lot of bug fixing work here we report result for https://github.com/src-d/style-analyzer/commit/eb2ec03d702cfbacfe2bfd1601cede2320ea4884
[quality-report-eb2ec03d702cfbacfe2bfd1601cede2320ea4884.tar.gz](https://github.com/src-d/style-analyzer/files/2864951/quality-report-eb2ec03d702cfbacfe2bfd1601cede2320ea4884.tar.gz)

PR #584, opened 2019-02-05
Remove global indentation from code generator

zurk:
Related to https://github.com/src-d/style-analyzer/issues/522

So we thought that this change helps us to simplify the code of `CodeGenerator` class. But as I find out it is not so easy because we still need quite a lot of logic to process indentation changes carefully. 

To simplify the code I should implement https://github.com/src-d/style-analyzer/issues/523 first. And this helps a lot to simplify `CodeGenerator`.

But this part can be merged now.

### DIFF
--- a/lookout/style/format/benchmarks/evaluate_smoke.py
+++ b/lookout/style/format/benchmarks/evaluate_smoke.py
@@ -20,7 +20,7 @@ from lookout.style.format.analyzer import FormatAnalyzer
 from lookout.style.format.code_generator import CodeGenerator
 from lookout.style.format.feature_extractor import FeatureExtractor
 from lookout.style.format.tests.test_analyzer_integration import TestAnalyzer
-from lookout.style.format.utils import flatten_dict, merge_dicts
+from lookout.style.format.utils import merge_dicts
 from lookout.style.format.virtual_node import VirtualNode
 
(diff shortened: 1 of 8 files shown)


lookout/style/format/benchmarks/evaluate_smoke.py
```diff
@@ -218,28 +218,25 @@ def calc_metrics(bad_style_code: str, correct_style_code: str, fe: Optional[Feat
...
-        misdetection, undetected, detected_wrong_fix, detected_correct_fix = \
-            calc_aligned_metrics(*align3(bad_style_code, correct_style_code, predicted_code))
-        losses[indentation] = {
-            "misdetection": misdetection,
-            "undetected": undetected,
-            "detected_wrong_fix": detected_wrong_fix,
-            "detected_correct_fix": detected_correct_fix,
-            "predicted_file": predicted_code,
-        }
+    if fe is not None:
```
vmarkovtsev: I see

```python
# TODO (zurk): make optional arguments non-optional.
```

How about fixing this TODO in this PR?

zurk: it is unrelated. I prefer to do it in another one after I finish with CodeGenerator

PR #589, opened 2019-02-07
Refactor FeatureExtractor.extract_features

vmarkovtsev:


lookout/style/format/feature_extractor.py
```diff
@@ -235,6 +235,74 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None) \
                  the corresponding `VirtualNode`-s and the parents mapping \
                  or None in case no features were extracted.
         """
+        parsed_files, node_parents, vnode_parents = self._parse_vnodes(files, lines)
+        xy = self._files_to_xy(parsed_files)
+        if xy is None:
+            return None
```
zurk: actually, I do not like we return None in this case. We either can return empty lists and dicts or raise an exception and catch it in a caller.

But I think you do not want to change feature extractor interface in this PR. So we can keep it for now. 

please tell me if you think it is ok as it is.

lookout/style/format/feature_extractor.py
```diff
@@ -235,6 +235,74 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None) \
                  the corresponding `VirtualNode`-s and the parents mapping \
                  or None in case no features were extracted.
         """
+        parsed_files, node_parents, vnode_parents = self._parse_vnodes(files, lines)
+        xy = self._files_to_xy(parsed_files)
```
zurk: should we always start function name with a verb? Consider to use `_convert_files_to_xy`.
I am fine with this name tho

lookout/style/format/feature_extractor.py
```diff
@@ -235,6 +235,74 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None) \
...
+        if xy is None:
+            return None
+        X, y, vnodes_y, vnodes, sibling_indices_list = xy
+        self._log.debug("Features shape: %s", X.shape)
+        self._log.debug("Labels shape: %s", y.shape)
```
zurk: let's remove it. It is obvious from the `X.shape`

lookout/style/format/feature_extractor.py
```diff
@@ -235,6 +235,74 @@ def extract_features(self, files: Iterable[File], lines: List[List[int]]=None) \
...
+            else:
+                self.selected_features = feature_selector.get_support(indices=True)
+        self._log.debug("Features shape after selection: %s" % (X.shape,))
+        if self.selected_features is None:
+            self.selected_features = numpy.arange(X.shape[1])
+        self._compute_feature_info()
+        return X, self.selected_features
+
+    @property
+    def composite_class_representations(self) -> List[str]:
```
zurk: should we place properties right after `__main__`?

lookout/style/format/feature_extractor.py
vmarkovtsev: Extracting zero vnodes is definitely not an error, so yep, we should return empty lists. I leave it for you :)

lookout/style/format/feature_extractor.py
vmarkovtsev: The first dimension yes, but the second?

lookout/style/format/feature_extractor.py
vmarkovtsev: I guess you mean `__init__`

vmarkovtsev:
@zurk properties reordered, function renamed

lookout/style/format/feature_extractor.py
zurk: it is 1-d array, so no second dimention

zurk:
https://github.com/src-d/style-analyzer/pull/589#discussion_r254732305

vmarkovtsev:
OK removed

Issue #606, opened 2019-02-14
FormatAnalyser should not fail if CodeGenerator fails


zurk:
In this case, we just miss Github Suggestion Feature but can at least give a message without it.
I suggest to log a warning about such cases and investigate them later without FormatAnalyser failure.

PR #614, opened 2019-02-15
Handle exceptions from CodeGenerator

zurk:
Closes https://github.com/src-d/style-analyzer/issues/606

### DIFF
--- a/lookout/style/format/analyzer.py
+++ b/lookout/style/format/analyzer.py
@@ -428,7 +428,15 @@ class FormatAnalyzer(Analyzer):
         for line_number, line in self._group_line_nodes(
                 y, y_pred, vnodes_y, new_vnodes, rule_winners):
             line_ys, line_ys_pred, line_vnodes_y, new_line_vnodes, line_winners = line
-            new_code_line = code_generator.generate_new_line(new_line_vnodes)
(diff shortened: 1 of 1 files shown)


lookout/style/format/analyzer.py
```diff
@@ -428,7 +428,15 @@ def _generate_token_fixes(
         for line_number, line in self._group_line_nodes(
                 y, y_pred, vnodes_y, new_vnodes, rule_winners):
             line_ys, line_ys_pred, line_vnodes_y, new_line_vnodes, line_winners = line
-            new_code_line = code_generator.generate_new_line(new_line_vnodes)
+            try:
+                new_code_line = code_generator.generate_new_line(new_line_vnodes)
+            except Exception as e:
```
vmarkovtsev:
```suggestion
            except:
```

zurk: No, Vadim, I should use `Exception` because you also have  `BaseException` as the highest level. I do not want to catch `SystemExit` `KeyboardInterupt` and `GeneratorExit`

vmarkovtsev:
```suggestion
            except Exception:
```

Issue #524, opened 2019-01-10
Use CodeGenerator for UAST filtering


vmarkovtsev:
Important

PR #621, opened 2019-02-18
Use CodeGenerator in uast check

zurk:
Closes https://github.com/src-d/style-analyzer/issues/524

It was hard to avoid any refactoring work because this change is hard to implement in the previous code.
So let me summarize what I did:
* One bug was fixed: if you have `-1` in `y_pred` it means we refuse to predict. We did not count it, so we consider it as the last label prediction. 
* `UASTStabilityChecker` class created. All main methods are inside. 
* `UASTStabilityChecker` has `check` and `check_file` methods. Before we have only `check` analog but often need to check only for one file
* `CodeGenerator.generate_one_change` special function for `UASTStabilityChecker`. It allows you to generate code for only one change in vnodes labels.
* minor bug in CodeGenerator: it is a bad idea to replace QUOTE with NOOP
*  debug mode to see refused changes like:
```
+++ suggested
@@ -79,8 +79,7 @@
       return null;
     }
 
-    // rgb/rgba(), hsl/hsla()
-    match = text.toLowerCase().match(/^\s*(?:(rgba?)|(hsla?))\((.*)\)\s*$/);
+    // rgb/rgba(), hsl/hsla() match = text.toLowerCase().match(/^\s*(?:(rgba?)|(hsla?))\((.*)\)\s*$/);
 
     if (match) {
       const components = match[3].trim();
```
^ yeah, deffenetely bad prediction. 

I run the quality report to see that everything is ok and it works faster and no bugs for first 4-5 repos.
Note that this change can affect our quality. To see how we can run report generation. 

@warenlg I know you are on vacations but if you have time to review I glad to have your comments.


lookout/style/format/postprocess.py
```diff
@@ -1,181 +1,265 @@
 """Postprocess predictions of the model."""
```
m09: since we create a new class and it's the only content of the file, we should rename the file `uast_stability_checker`

lookout/style/format/postprocess.py
```diff
@@ -1,181 +1,265 @@
...
 
 
-_log = getLogger("Postprocess")
-
-
-def check_uasts_are_equal(uast1: bblfsh.Node, uast2: bblfsh.Node) -> bool:
+class UASTStabilityChecker:
     """
-    Check if 2 UASTs are identical or not in terms of nodes `roles`, `internal_type` and `token`.
+    Check if predictions changes UAST structure of the processing files.
```
m09:
```suggestion
    Check if predictions change the UAST structure of the processed files.
```

lookout/style/format/postprocess.py
```diff
@@ -1,181 +1,265 @@
...
+                break
+            current_ancestor = node_parents.get(id(current_ancestor), None)
+        else:
+            result = None
+            self._log.warning("skipped file %s, due to errors in parsing the whole content", path)
+        for descendant in descendants:
+            self._parsing_cache[id(descendant)] = result
+        return result
+
+    def check_file(
```
m09: We should keep this method private (`_check_file`) and use `[file]` as the `files` argument of `check` and keep the API simple, at least we should provide a nice docstring explaning the difference between the two (atm it's the same docstring so diving into the code is required).

lookout/style/format/postprocess.py
```diff
@@ -1,181 +1,265 @@
...
+        return tuple(res)
+
+    @staticmethod
+    def check_uasts_are_equal(uast1: bblfsh.Node, uast2: bblfsh.Node) -> bool:
+        """
+        Check if 2 UASTs are identical in terms of nodes `roles`, `internal_type` and `token`.
+
+        :param uast1: The bblfsh.Node of the first UAST to compare.
+        :param uast2: The bblfsh.Node of the second UAST to compare.
+        :return: A boolean equals to True if the 2 input UASTs are identical and False otherwise.
```
m09:
```suggestion
        :return: True if the 2 input UASTs are identical and False otherwise.
```

lookout/style/format/tests/test_postprocess.py
```diff
@@ -82,12 +82,37 @@ def test_posprocess(self):
     def test_bad_and_good_quotes(self):
         self.edit_and_test("""var a = '"0"'; var c = "0";""",
                            {4: (CLS_DOUBLE_QUOTE), 5: (CLS_DOUBLE_QUOTE,),
-                            10: (CLS_SINGLE_QUOTE), 11: (CLS_SINGLE_QUOTE,)},
+                            10: (CLS_SINGLE_QUOTE), 11: (CLS_SINGLE_QUOTE)},
```
m09: I see that tuples were changed in the PR splitting quotes and other tokens but `_to_label` expects sequences of labels. It still works because technically a string is a sequence so even a character is a sequence but I'd rather keep the tuples. So basically:

                               {4: (CLS_DOUBLE_QUOTE,), 5: (CLS_DOUBLE_QUOTE,),
                                10: (CLS_SINGLE_QUOTE,), 11: (CLS_SINGLE_QUOTE,)},

zurk: good catch! thanks.

lookout/style/format/postprocess.py
zurk: Good point, I think I found a good solution. See new changes.

zurk:
> test suite to QA this method

@m09 do you mean that it is some code not in the repository?

m09:
> > test suite to QA this method
> 
> @m09 do you mean that it is some code not in the repository?

I don't know if it's in the research folder or not. I think it's mainly that @warenlg runs the uast check against a known dataset and has previous results somewhere to check the drop/increase in performance and some specific cases. I don't think it's worth including as a full report but indeed having the procedure in research could be beneficial.

lookout/style/format/uast_stability_checker.py
```diff
@@ -0,0 +1,253 @@
...
+        :param stub: Babelfish GRPC service stub.
+        :param vnode_parents: `VirtualNode`-s' parents mapping as the LCA of the closest
+                               left and right babelfish nodes.
+        :param node_parents: Parents mapping of the input UASTs.
+        :param rule_winners: Numpy array of the index of the winning rule for each sample.
+        :param grouped_quote_predictions: Quotes predictions (handled differenlty from the rest).
+        :return: List of predictions indices that are considered valid i.e. that are not breaking
+                 the UAST.
+        """
+        if isinstance(files, File):
```
m09: I'm not against this optimization (the other way being to just apply the splitting code in the case of a single file and pay the small overhead for the sake of simplicity) but I'd rather keep a list in all cases and have a test for len == 1. This way we don't encounter the counter-intuitive `files=file` and we avoid union in types which keeps mypy useful.

lookout/style/format/uast_stability_checker.py
```diff
@@ -0,0 +1,253 @@
...
+        """
+        Filter the model's predictions that modify the UAST apart from changing Node positions.
+
+        :param y: Numpy 1-dimensional array of labels.
+        :param y_pred: Numpy 1-dimensional array of predicted labels by the model.
+        :param vnodes_y: Sequence of the labeled `VirtualNode`-s corresponding to labeled samples.
+        :param vnodes: Sequence of all the `VirtualNode`-s corresponding to the input.
+        :param files: File or Sequence of File-s with content, uast and path.
+        :param stub: Babelfish GRPC service stub.
+        :param vnode_parents: `VirtualNode`-s' parents mapping as the LCA of the closest
```
vmarkovtsev:
```suggestion
        :param vnode_parents: `VirtualNode`-s' parents mapping as the LCA of the closest \
```

lookout/style/format/uast_stability_checker.py
```diff
@@ -0,0 +1,253 @@
...
+                               left and right babelfish nodes.
+        :param node_parents: Parents mapping of the input UASTs.
+        :param rule_winners: Numpy array of the index of the winning rule for each sample.
+        :param grouped_quote_predictions: Quotes predictions (handled differenlty from the rest).
+        :return: List of predictions indices that are considered valid i.e. that are not breaking
```
vmarkovtsev:
```suggestion
        :return: List of predictions indices that are considered valid i.e. that are not breaking \
```

lookout/style/format/uast_stability_checker.py
```diff
@@ -0,0 +1,253 @@
...
+        :param y: Numpy 1-dimensional array of labels.
+        :param y_pred: Numpy 1-dimensional array of predicted labels by the model.
+        :param vnodes_y: Sequence of the labeled `VirtualNode`-s corresponding to labeled samples.
+        :param vnodes: Sequence of all the `VirtualNode`-s corresponding to the input.
+        :param files: File or Sequence of File-s with content, uast and path.
+        :param stub: Babelfish GRPC service stub.
+        :param vnode_parents: `VirtualNode`-s' parents mapping as the LCA of the closest
+                               left and right babelfish nodes.
+        :param node_parents: Parents mapping of the input UASTs.
+        :param rule_winners: Numpy array of the index of the winning rule for each sample.
```
vmarkovtsev:
```suggestion
        :param rule_winners: Numpy array with the indexes of the winning rules for each sample.
```

lookout/style/format/uast_stability_checker.py
vmarkovtsev: This is confusing, yes

lookout/style/format/uast_stability_checker.py
```diff
@@ -0,0 +1,253 @@
...
+        for res_i in zip(*check_result):
+            if isinstance(res_i[0], list):
+                res.append(list(chain(*res_i)))
+            else:
+                res.append(numpy.concatenate(res_i))
+            pass
+        return tuple(res)
+
+    @staticmethod
+    def check_uasts_are_equal(uast1: bblfsh.Node, uast2: bblfsh.Node) -> bool:
```
vmarkovtsev:
```suggestion
    def check_uasts_are_equivalent(uast1: bblfsh.Node, uast2: bblfsh.Node) -> bool:
```

lookout/style/format/uast_stability_checker.py
```diff
@@ -0,0 +1,253 @@
...
+        """
+        Check if 2 UASTs are identical in terms of nodes `roles`, `internal_type` and `token`.
```
vmarkovtsev:
```suggestion
        Check if 2 UAST nodes are identical regarding `roles`, `internal_type` and `token` of their subtree members.
```

lookout/style/format/uast_stability_checker.py
```diff
@@ -0,0 +1,253 @@
...
+        for descendant in descendants:
+            self._parsing_cache[id(descendant)] = result
+        return result
+
+    def _check_file(
+            self, y: numpy.ndarray, y_pred: numpy.ndarray, vnodes_y: Sequence[VirtualNode],
+            vnodes: Sequence[VirtualNode], file: File, stub: "bblfsh.aliases.ProtocolServiceStub",
+            vnode_parents: Mapping[int, bblfsh.Node], node_parents: Mapping[int, bblfsh.Node],
+            rule_winners: numpy.ndarray, grouped_quote_predictions: QuotedNodeTripleMapping,
+    ) -> _check_return_type:
```
vmarkovtsev: Missing: current algorithm description (can be TODO for Waren) and also

```python
# TODO(vmarkovtsev): apply ML to not parse all the parents
```

lookout/style/format/uast_stability_checker.py
```diff
@@ -0,0 +1,253 @@
...
+    _check_return_type = Tuple[numpy.ndarray, numpy.ndarray, Sequence[VirtualNode], numpy.ndarray,
+                               numpy.ndarray]
+
+    def __init__(self, feature_extractor: FeatureExtractor, debug: bool=False):
+        """
+        Construct a UASTStabilityChecker.
+
+        :param feature_extractor: Feature extraction class that was used to generate data for \
+                                  the check.
+        :param debug: Logs code diff for unsafe predictions. on debug level.
```
vmarkovtsev:
```suggestion
        :param debug: Logs code diff for unsafe predictions with debug level.
```

lookout/style/format/uast_stability_checker.py
zurk: let's omit `are`: `check_uasts_equivalent`?

lookout/style/format/code_generator.py
```diff
@@ -130,6 +130,22 @@ def apply_predicted_y(self, vnodes: Sequence[VirtualNode],
             result.append(vnode)
         return result
 
+    def generate_one_change(self, vnodes: Sequence[VirtualNode], changed_vnode_index: int,
+                            new_label: Tuple[int, ...]) -> str:
+        """
+        Generate new source code for only one change.
```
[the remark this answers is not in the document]
zurk: clarified in the doc.

lookout/style/format/uast_stability_checker.py
vmarkovtsev: No objections

m09:
@warenlg we'll PR new changes if we find out that there is a drop in performance that we missed.

PR #615, opened 2019-02-15
Refactor Columns naming logic.

irinakhismatullina:
Change columns' names interchange logic.

Also there is some very small refactoring present. Sorry, @vmarkovtsev, in the end it was difficult to detach one from the other.

PR #623, opened 2019-02-19
Refactor typos columns naming

irinakhismatullina:
First part of #615


lookout/style/typos/utils.py
```diff
@@ -2,22 +2,19 @@
...
-SPLIT_COLUMN = "token_split"
-SUGGESTIONS_COLUMN = "suggestions"
-TYPO_COLUMN = "typo"
+Columns = NamedTuple(
+    "Columns",
+    [("Token", str), ("CorrectToken", str), ("Split", str), ("CorrectSplit", str), ("After", str),
+     ("Before", str), ("Id", str), ("Candidate", str), ("Features", str), ("Probability", str),
+     ("Suggestions", str)])(
+    "token", "correct_token", "token_split", "correct_token_split", "after", "before", "id",
+    "candidate", "features", "proba", "suggestions")
```
zurk: Why do you use `NamedTuple`?

I think it is better to use Enum in such case. 
```python
from enum import Enum, unique


@unique
class Columns(Enum):
    """
    DOC HERE
    """
    Token = "token"
    CorrectToken = "correct_token"
    ...
```

irinakhismatullina: Because this way I will have to write, for example, `Columns.Token.value` instead of `Columns.Token`. Looks like Enum with string fields can only work like that, I checked.

PR #624, opened 2019-02-19
Refactor corrector

irinakhismatullina:
1. Merge `train()` with `train_on_file()`, `suggest()` with `suggest_file()`.
2. Add threads_number setter.
3. Fix docstrings.

Based on #623. Second part of #615.


lookout/style/typos/corrector.py
```diff
@@ -49,10 +49,15 @@ def __init__(self, **kwargs):
         self.ranker = CandidatesRanker()
 
     @property
-    def threads_number(self):
-        """Return the number of threads used to train and to predict."""
+    def threads_number(self) -> int:
+        """Return the number of threads for multiprocessing used to train and to predict."""
         return self.ranker.boost_param["nthread"]
 
+    @threads_number.setter
```
zurk: I do not see where do you use it. why do you need this setter?

lookout/style/typos/corrector.py
```diff
@@ -77,118 +82,88 @@ def initialize_generator(self, vocabulary_file: str, frequencies_file: str,
...
-        if candidates is None:
-            candidates = self.generator.generate_candidates(typos, self.threads_number,
-                                                            save_candidates_file)
+        if isinstance(data, str):
+            data = pandas.read_csv(data, index_col=0)
+        if not candidates:
+            candidates = self.generator.generate_candidates(
+                data, self.threads_number, save_candidates_file)
+        elif isinstance(candidates, str):
+            candidates = pandas.read_pickle(candidates)
```
zurk: why do you use not only csv files for everything?

lookout/style/typos/corrector.py
```diff
@@ -77,118 +82,88 @@ def initialize_generator(self, vocabulary_file: str, frequencies_file: str,
...
+            data = pandas.read_csv(data, index_col=0)
         all_suggestions = []
-        for i in tqdm(range(0, len(typos), batch_size)):
-            suggestions = self.suggest(typos.loc[typos.index[i]:
-                                                 typos.index[min(len(typos) - 1,
-                                                                 i + batch_size - 1)], :],
-                                       n_candidates=n_candidates, return_all=return_all)
+        for i in tqdm(range(0, len(data), batch_size)):
+            suggestions = self.suggest(
+                data.loc[data.index[i]:data.index[min(len(data) - 1, i + batch_size - 1)], :],
```
zurk: maybe 
```python
data.loc[data.index[i]:data.index[min(len(data), i + batch_size)], :],
```
For examnple
```python
l = [0, 1, 2, 3, 4, 5, 6]
l[1:4]
```
gives you `[1, 2, 3]` without `4`

lookout/style/typos/corrector.py
irinakhismatullina: It is a way to set number of threads for multiprocessing. It should be provided. And it will definitely be used in upcoming code.

lookout/style/typos/corrector.py
irinakhismatullina: Well, for historical reasons:) Also `candidates` df is usually uite massive, several times bigger than `data`. But I'm thinking about moving to .csv everywhere myself.

lookout/style/typos/corrector.py
irinakhismatullina: Because apparantly pandas.DataFrame.loc includes both ends (I was surprised too):

> Slice with integer labels for rows. As mentioned above, note that both the start and stop of the slice are included.

from https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.loc.html

vmarkovtsev: Wow this is unexpected at all

vmarkovtsev:
@irinakhismatullina Can you please rebase this once since #623 has been merged.

lookout/style/typos/corrector.py
```diff
@@ -77,118 +82,88 @@ def initialize_generator(self, vocabulary_file: str, frequencies_file: str,
...
+        Train corrector on the given dataset of typoed tokens inside identifiers.
 
-        :param typos: DataFrame containing columns "typo" and "identifier",
-                      column "token_split" is optional, but used when present.
-        :param candidates: DataFrame with precalculated candidates.
-        :param save_candidates_file: Path to file where to save the candidates.
+        :param data: DataFrame or its .csv dump, containing columns Columns.Token and \
+                     Columns.CorrectToken, column Columns.Split is optional, but used when present.
+        :param candidates: DataFrame or its pickle dump with precalculated candidates.
+        :param save_candidates_file: Path to file where to save the candidates (.pkl).
```
vmarkovtsev: Our convention is to use `.pickle`

lookout/style/typos/corrector.py
```diff
@@ -77,118 +82,88 @@ def initialize_generator(self, vocabulary_file: str, frequencies_file: str,
...
                                 equal edit distance from the typo to consider as candidates.
         :param max_distance: Maximum edit distance for symspell lookup.
         :param radius: Maximum edit distance from typo allowed for candidates.
-        :return: Nothing
         """
         self.generator.construct(vocabulary_file, frequencies_file, embeddings_file,
                                  neighbors_number, edit_candidates, max_distance, radius)
 
-    def train(self, typos: pandas.DataFrame, candidates: pandas.DataFrame = None,
+    def train(self, data: Union[pandas.DataFrame, str],
```
vmarkovtsev: I would prefer leaving only `DataFrame` parameters. The loading from CSV can be performed in a separate convenience function.

lookout/style/typos/corrector.py
```diff
@@ -77,118 +82,88 @@ def initialize_generator(self, vocabulary_file: str, frequencies_file: str,
...
-               column "token_split" is optional, but used when present
+        :param data: DataFrame or its .csv dump, containing column "typo", \
+                     column "token_split" is optional, but used when present
         :param n_candidates: Number of most probable candidates to return
         :param return_all: False to return suggestions only for corrected tokens
         :param batch_size: Batch size
         :return: Dictionary {id : [[candidate, correctness_proba]]}, candidates are sorted \
                  by correctness probability in a descending order.
         """
+        if isinstance(data, str):
```
vmarkovtsev: same here

lookout/style/typos/corrector.py
zurk: just it looks like it is related to logic, but not to a model. 
If you need this number in your Analyser it is better to store it in analyzer config.

lookout/style/typos/corrector.py
zurk: is it possible to implement in this pr? 
you can use `dataframe.to_csv(..., compression="xz")` to decrease the file size.

lookout/style/typos/corrector.py
zurk: Woooooow

zurk: can you place a comment with the explanation that it, as is, should be and a link to the doc, please?

lookout/style/typos/corrector.py
vmarkovtsev: This is actually a nice suggestion @zurk

Then it is possible to use xzcat/xzless to see inside.

lookout/style/typos/corrector.py
irinakhismatullina: Well, anyone who uses the model can need to set this number, not only analyzer. It is not the thing that we store, but the thing that is important for running the mode. I still think this method should stay. @vmarkovtsev, could you give your opinion?

lookout/style/typos/corrector.py
irinakhismatullina: This PR is all about removing those separate methods for using data from files. So should I return them back, or there must be some external wrapping?

lookout/style/typos/corrector.py
irinakhismatullina: Cool, thank you, I will do that.

lookout/style/typos/corrector.py
irinakhismatullina: Ok

m09: Maybe you need `iloc`? `loc` is label based. With `iloc` I'd write

    data.iloc[i:i + batch_size, :]
I didn't read the rest of the code so there is maybe a good reason to use `loc` I'm mentioning it just in case.

irinakhismatullina: Yes, it's much better this way, thank you!

lookout/style/typos/corrector.py
vmarkovtsev: I don't see anything bad in this setter as long as it is tested. If we see while preparing the release that this code is indeed not used, we are going to delete it anyway.

lookout/style/typos/corrector.py
```diff
@@ -77,118 +82,110 @@ def initialize_generator(self, vocabulary_file: str, frequencies_file: str,
...
-                      column "token_split" is optional, but used when present.
-        :param candidates: DataFrame with precalculated candidates.
-        :param save_candidates_file: Path to file where to save the candidates.
+        :param data: DataFrame, containing columns Columns.Token and Columns.CorrectToken, \
+                     column Columns.Split is optional, but used when present.
+        :param candidates: A .csv.xz dump of a dataframe with precalculated candidates.
+        :param save_candidates_file: Path to file where to save the candidates (.csv.xz).
         """
-        if candidates is None:
+        if not candidates:
```
vmarkovtsev: The default value is `None` so we should check `None`. An empty string may signal about a problem up the stack so we should consider it as a legitimate value.

lookout/style/typos/corrector.py
```diff
@@ -77,118 +82,110 @@ def initialize_generator(self, vocabulary_file: str, frequencies_file: str,
...
-        Suggest corrections for given typos.
+        Suggest corrections for the tokens from the given dataset.
 
-        :param typos: DataFrame containing column "typo", \
-                      column "token_split" is optional, but used when present
-        :param candidates: DataFrame with precalculated candidates
-        :param n_candidates: Number of most probable candidates to return
-        :param return_all: False to return suggestions only for corrected tokens
-        :param save_candidates_file: Path to file to save candidates to
+        :param data: DataFrame, containing column Columns.Token, column Columns.Split \
```
vmarkovtsev:
```suggestion
        :param data: DataFrame which contains Columns.Token. Columns.Split \
```

lookout/style/typos/corrector.py
```diff
@@ -77,118 +82,110 @@ def initialize_generator(self, vocabulary_file: str, frequencies_file: str,
...
+        :param save_candidates_file: Path to file to save candidates to (.csv.xz).
+        :param n_candidates: Number of most probable candidates to return.
+        :param return_all: False to return suggestions only for corrected tokens.
         :return: Dictionary {id : [[candidate, correctness_proba]]}, candidates are sorted \
                  by correctness probability in a descending order.
         """
-        if candidates is None:
-            candidates = self.generator.generate_candidates(typos, self.threads_number,
-                                                            save_candidates_file)
+        if not candidates:
```
vmarkovtsev: Again, we should check `None`

lookout/style/typos/corrector.py
```diff
@@ -77,118 +82,110 @@ def initialize_generator(self, vocabulary_file: str, frequencies_file: str,
...
+        Suggest corrections for the tokens from the given dataset.
 
-        :param typos: DataFrame containing column "typo", \
-                      column "token_split" is optional, but used when present
-        :param candidates: DataFrame with precalculated candidates
-        :param n_candidates: Number of most probable candidates to return
-        :param return_all: False to return suggestions only for corrected tokens
-        :param save_candidates_file: Path to file to save candidates to
+        :param data: DataFrame, containing column Columns.Token, column Columns.Split \
+                     is optional, but used when present.
```
vmarkovtsev: I really don't like this optionality. Do we use it in our normal analysis flow? If we do, make it required. If we don't, remove it.

lookout/style/typos/corrector.py
zurk: ok, let's keep it.

lookout/style/typos/corrector.py
vmarkovtsev: It is better to have separate methods to load and process the data, yes. One methods loads from csv (or whatever) to dataframe and the other works with this dataframe.

PR #630, opened 2019-02-20
Add filter_splits preprocesing function

irinakhismatullina:
Continue with functions for preprocessing raw data.

This one filters identifiers by leaving only those, whose tokens are all in the vocabulary. This filtered data is then used for creating train and test datasets with artificially made typos.

### DIFF
--- a/lookout/style/typos/preprocessing.py
+++ b/lookout/style/typos/preprocessing.py
@@ -1,4 +1,4 @@
-from typing import Set, Union
+from typing import Set
 
 import pandas
 
@@ -19,14 +19,12 @@ def check_split(text: str, tokens: Set[str]) -> bool:
     return True
 
 
-def filter_splits(data: Union[pandas.DataFrame, str], tokens: Set[str]) -> pandas.DataFrame:
+def filter_splits(data: pandas.DataFrame, tokens: Set[str]) -> pandas.DataFrame:
     """
     Leave rows in a dataframe whose splits' tokens all belong to some vocabulary.
 
-    :param data: Dataframe or its .csv dump, containing column Columns.Split.
+    :param data: Dataframe which contains column Columns.Split.
     :param tokens: Set of tokens (reference vocabulary).
     :return: Filtered dataframe.
     """
-    if isinstance(data, str):
-        data = pandas.read_csv(data, index_col=0)
-    return data.loc[data[Columns.Split].apply(lambda x: check_split(x, tokens))]
+    return data[data[Columns.Split].apply(lambda x: check_split(x, tokens))]
--- a/lookout/style/typos/preprocessing.py
+++ b/lookout/style/typos/preprocessing.py
@@ -5,31 +5,28 @@ import pandas
 from lookout.style.typos.utils import Columns
 
 
-def check_split(split: str, tokens_set: Set[str]) -> bool:
+def check_split(text: str, tokens: Set[str]) -> bool:
     """
-    Check whether all elements of the split are in tokens_set.
+    Check whether all tokens of the text belong to the given set.
 
-    :param split: String of space-separated tokens.
-    :param tokens_set: Set of tokens (reference vocabulary).
-    :return: True when all tokens of split belong to tokens_set.
+    :param text: String of space-separated tokens.
+    :param tokens: Set of tokens (reference vocabulary).
+    :return: True when all tokens of the `text` belong to `tokens`.
     """
-    if type(split) != str:
-        return False
-
-    for token in split.split():
-        if token not in tokens_set:
+    for token in text.split():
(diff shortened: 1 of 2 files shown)


lookout/style/typos/preprocessing.py
```diff
@@ -0,0 +1,35 @@
...
+
+def check_split(split: str, tokens_set: Set[str]) -> bool:
+    """
+    Check whether all elements of the split are in tokens_set.
+
+    :param split: String of space-separated tokens.
+    :param tokens_set: Set of tokens (reference vocabulary).
+    :return: True when all tokens of split belong to tokens_set.
+    """
+    if type(split) != str:
```
zurk:
```suggestion
    if not isinstance(str, split):
```
But you specify the type as str in function header: `split: str` so that means that you do not expect other types. 
You should remove this condition or fix type annotation.

lookout/style/typos/preprocessing.py
```diff
@@ -0,0 +1,35 @@
+from typing import Set, Union
+
+import pandas
+
+from lookout.style.typos.utils import Columns
+
+
+def check_split(split: str, tokens_set: Set[str]) -> bool:
```
zurk: `split.split()` looks strange to me.  consider renaming `split` to something else. For example, if it is a text line you can call it `line`, or just `text` or at least `to_split`

zurk: Also, we do not use variable type in its name usually, so `tokens` totally file here I think. You do not have a list of tokens or any other structure.

lookout/style/typos/preprocessing.py
```diff
@@ -0,0 +1,35 @@
...
+    if type(split) != str:
+        return False
+
+    for token in split.split():
+        if token not in tokens_set:
+            return False
+    return True
+
+
+def filter_splits(data: Union[pandas.DataFrame, str], tokens_set: Set[str]) -> pandas.DataFrame:
```
zurk: `token_set` -> `token`

And data type is just `pandas.DataFrame` as I can see.

lookout/style/typos/preprocessing.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+    Leave rows in a dataframe whose splits' tokens all belong to some vocabulary.
+
+    :param data: Dataframe or its .csv dump, containing column Columns.Split.
```
zurk:
```suggestion
    :param data: Dataframe containing column Columns.Split.
```

lookout/style/typos/preprocessing.py
```diff
@@ -0,0 +1,35 @@
...
+    :param tokens_set: Set of tokens (reference vocabulary).
+    :return: Filtered dataframe.
+    """
+    token_split = list(data[Columns.Split])
+    filter_array = list(map(lambda x: check_split(x, tokens_set), token_split))
+    return data[filter_array]
```
zurk: it is always better to come up with some vectorized code. I did not check it, but something like
```python
return data.loc[data[Columns.Split].map(lambda x: check_split(x, tokens_set))]
```

should work

lookout/style/typos/preprocessing.py
```diff
@@ -0,0 +1,32 @@
...
+
+def filter_splits(data: Union[pandas.DataFrame, str], tokens: Set[str]) -> pandas.DataFrame:
+    """
+    Leave rows in a dataframe whose splits' tokens all belong to some vocabulary.
+
+    :param data: Dataframe or its .csv dump, containing column Columns.Split.
+    :param tokens: Set of tokens (reference vocabulary).
+    :return: Filtered dataframe.
+    """
+    if isinstance(data, str):
```
m09: As @vmarkovtsev mentioned in another PR, it's important to split the IO part and the logic part (to have the biggest possible part of the codebase made of [pure functions](https://en.wikipedia.org/wiki/Pure_function)). This enhances testability and maintainability by a lot.

We should therefore remove this check.

lookout/style/typos/preprocessing.py
```diff
@@ -0,0 +1,32 @@
...
+    :param tokens: Set of tokens (reference vocabulary).
+    :return: True when all tokens of the `text` belong to `tokens`.
+    """
+    for token in text.split():
+        if token not in tokens:
+            return False
+    return True
+
+
+def filter_splits(data: Union[pandas.DataFrame, str], tokens: Set[str]) -> pandas.DataFrame:
```
m09: and adapt this type to be only `pandas.DataFrame`

lookout/style/typos/preprocessing.py
irinakhismatullina: Very well, will do that everywhere.

PR #636, opened 2019-02-22
Fix and simplify summary report output&save

zurk:


lookout/style/format/benchmarks/quality_report.py
```diff
@@ -430,44 +429,41 @@ def generate_quality_report(input: str, output: str, force: bool, bblfsh: str, t
...
+                    floatfmts.append(FLOAT_PRECISION)
+                elif field == "Average Rule Len":
+                    floatfmts.append(".1f")
+                else:
+                    floatfmts.append("g")
 
-                print(tabulate(table, tablefmt="pipe", headers="firstrow", floatfmt=floatfmts),
-                      file=output)
-                summary = output.getvalue()
+            summary = tabulate(table, tablefmt="pipe", headers="firstrow", floatfmt=floatfmts)
```
vmarkovtsev: May I ask you to improve this even further: return the string table from the function.

Then in the higher level code:

```python
summary = generate_quality_report(...)
log.info("%s", report_name)
log.info("\n%s", summary)
summary_loc = os.path.join(output, "summary-%s.md" % report_name)
with open(summary_loc, "w", encoding="utf-8") as f:
    f.write(summary)
```

We can put it to another function, say `write_quality_report`

zurk: good point.

lookout/style/format/benchmarks/quality_report.py
```diff
@@ -321,8 +320,51 @@ def handle_input_arg(input_arg: str, log: Optional[logging.Logger] = None) -> It
...
+            floatfmts.append(".1f")
+        else:
+            floatfmts.append("g")
+
+    return tabulate(table, tablefmt="pipe", headers="firstrow", floatfmt=floatfmts)
+
+
 def generate_quality_report(input: str, output: str, force: bool, bblfsh: str, train_config: dict,
-                            database: Optional[str] = None, fs: Optional[str] = None) -> None:
+                            database: Optional[str] = None, fs: Optional[str] = None) -> str:
```
vmarkovtsev: We don't return a string now

lookout/style/format/benchmarks/quality_report.py
```diff
@@ -425,49 +467,8 @@ def generate_quality_report(input: str, output: str, force: bool, bblfsh: str, t
...
-                    elif field == "Average Rule Len":
-                        floatfmts.append(".1f")
-                    else:
-                        floatfmts.append("g")
-
-                print(tabulate(table, tablefmt="pipe", headers="firstrow", floatfmt=floatfmts),
-                      file=output)
-                summary = output.getvalue()
+            summary = _generate_report_summary(reports, report_name)
             print(report_name)
```
vmarkovtsev: Can you please change these two print-s to logging (as I suggested in the original snippet)

zurk: I can but I think we can consider `summary` as an output result and print it to sdtin to make all Unix pipe thing work. 
WDYT?

vmarkovtsev: I would agree with you if we did not write the same data to a file at the same time.

lookout/style/format/benchmarks/quality_report.py
```diff
@@ -425,51 +467,9 @@ def generate_quality_report(input: str, output: str, force: bool, bblfsh: str, t
...
-                    else:
-                        floatfmts.append("g")
-
-                print(tabulate(table, tablefmt="pipe", headers="firstrow", floatfmt=floatfmts),
-                      file=output)
-                summary = output.getvalue()
-            print(report_name)
-            print(summary)
+            summary = _generate_report_summary(reports, report_name)
+            log.info("\n%s\n%s", report_name, summary)
```
vmarkovtsev: Question: how does it work if we don't have the `log` argument

zurk: We have it
https://github.com/src-d/style-analyzer/pull/636/files#diff-ab1c29cb3f4a8283ca131cb3073b7f51R384

zurk: I run report just to check that everything is fine. everything is fine :)

PR #639, opened 2019-02-22
Fix FormatAnalyzer.check_training_required

vmarkovtsev:


vmarkovtsev:
This is no longer needed.

PR #640, opened 2019-02-22
Fix the FormatModel

zurk:


vmarkovtsev:
Forgot to update `container/Dockerfile` @zurk

zurk:
@vmarkovtsev maybe we want to use https://docs.docker.com/engine/reference/builder/#arg? 
it also required to customize `make docker-build` in `Makefile` but I think we can do that.

vmarkovtsev:
This Dockerfile always hardcodes the version of the upstream dependency - this is just they way it works. The same as you compose requirements.txt

PR #647, opened 2019-02-26
Move merge_dicts to common style utils

irinakhismatullina:
It is a preparation for using this function in typos corrector.

### DIFF
--- a/lookout/style/format/utils.py
+++ b/lookout/style/format/utils.py
@@ -1,6 +1,5 @@
 """Commonly used utils."""
-from copy import deepcopy
-from typing import Any, Dict, Iterable, Mapping, Sequence
+from typing import Any, Dict, Iterable, Sequence
 import warnings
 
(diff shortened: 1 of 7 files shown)


lookout/style/format/utils.py
```diff
@@ -78,38 +77,6 @@ def prepare_data_stub(input_pattern: str, client: BblfshClient, language: str):
...
-        d1, d2 = stack.pop()
-        for key, value in d2.items():
-            if isinstance(value, dict):
-                stack.append((d1.setdefault(key, {}), value))
-            else:
-                d1[key] = value
-    return res
-
-
 def generate_comment(filename: str, confidence: int, line: int, text: str) -> Comment:
```
zurk: let's move `generate_comment` also because it is common too.

vmarkovtsev: ...in a future PR because the scope is well-defined here

PR #649, opened 2019-02-26
Add prepare_data function and tests

irinakhismatullina:
The function comprises vocabulary and "clean" dataset (which contains only identifiers with tokens from the vocabulary).


lookout/style/typos/general.py
```diff
@@ -0,0 +1,83 @@
+from typing import Any, Mapping, Optional
```
vmarkovtsev: "general.py" does not seem to be a good module name. As of now, it looks like "train_dataset_preparation.py"

lookout/style/typos/general.py
```diff
@@ -0,0 +1,83 @@
...
+
+from google_drive_downloader import GoogleDriveDownloader as gdd
+import pandas
+
+from lookout.style.common import merge_dicts
+from lookout.style.typos.preprocessing import filter_splits, print_frequencies
+from lookout.style.typos.utils import Columns, flatten_df_by_column
+
+
+DRIVE_DATASET_ID = "1muNVWPe68XK8SFvqIv3V728NmkT46aTx"
```
vmarkovtsev: These should be the defaults overridable in function arguments. `DEFAULT_DRIVE_DATASET_ID`

lookout/style/typos/general.py
```diff
@@ -0,0 +1,83 @@
...
+DRIVE_FASTTEXT_ID = "1hCOIwKn-QZLVv1S385HxyNMeERgKGIvo"
```
vmarkovtsev: Fasttest vectors must be trained from scratch (they are not generic, are they?), we may not download it.

lookout/style/typos/general.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+defaults_for_preparation = {
+    "load_from_drive": False,
+    "input_path": "lookout/style/typos/data/raw_data.csv",
```
vmarkovtsev: `os.path.join` and relative to `__file__`

lookout/style/typos/general.py
```diff
@@ -0,0 +1,83 @@
...
+    "vocabulary_size": 10000,
+    "frequencies_size": None,
+    "vocabulary_path": "lookout/style/typos/data/vocabulary.csv",
+    "frequencies_path": "lookout/style/typos/data/frequencies.csv",
+}
+
+
+def prepare_data(params: Optional[Mapping[str, Any]] = None) -> pandas.DataFrame:
+    """
+    Get all necessary data from raw dataset of splitted identifiers with some statistics.
```
vmarkovtsev:
```suggestion
    Generate all the necessary data from the raw dataset of split identifiers.
```

lookout/style/typos/general.py
```diff
@@ -0,0 +1,83 @@
...
+                   frequency_column: Name of column with identifiers frequencies. If not \
+                                     specified, every split is considered to have frequency 1.
+                   vocabulary_size: Number of most frequent tokens to take as a vocabulary.
+                   frequencies_size: Number of most frequent tokens to save  frequencies info for.\
+                                     This information will be used by corrector as features for \
+                                     these tokens when they will be checked. If not specified, \
+                                     frequencies for all present tokens will be saved.
+                   vocabulary_path: .csv path to save vocabulary to.
+                   frequencies_path: .csv path to save frequencies to.
+    :return: Filtered dataset.
```
vmarkovtsev:
```suggestion
    :return: Dataset baked for training the typos correction.
```

lookout/style/typos/general.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    params = merge_dicts(defaults_for_preparation, params)
+    if params["load_from_drive"]:
+        gdd.download_file_from_google_drive(file_id=DRIVE_DATASET_ID,
```
vmarkovtsev: Normally you don't have this `load_from_drive` setting. Instead, you cache the downloaded data to a predefined file and check whether it exists.

lookout/style/typos/general.py
```diff
@@ -0,0 +1,83 @@
...
+                                            dest_path=params["input_path"],
+                                            overwrite=True)
```
vmarkovtsev: Would be great to have a progress bar here. Is it possible?

lookout/style/typos/tests/test_general_functions.py
```diff
@@ -0,0 +1,31 @@
...
+
+from lookout.style.typos.general import prepare_data
+from lookout.style.typos.utils import Columns, read_frequencies, read_vocabulary
+
+
+class PreparationTest(unittest.TestCase):
+    def test_prepare_data(self):
+        with tempfile.NamedTemporaryFile() as input_file:
+            with tempfile.NamedTemporaryFile() as voc_file:
+                with tempfile.NamedTemporaryFile() as freq_file:
```
zurk: Let's use 
```python

with tempfile.NamedTemporaryFile() as input_file, \
     tempfile.NamedTemporaryFile() as voc_file, \
     tempfile.NamedTemporaryFile() as freq_file:
```

lookout/style/typos/general.py
irinakhismatullina: Ok, then I will add that training to this function (it is also done on the raw data). Is it better to do it in this PR, or the next one?

vmarkovtsev: Better in a separate one

lookout/style/typos/general.py
irinakhismatullina: I tried this, but this way you can't store this raw data in temporary file, for example in tests.

vmarkovtsev: Add another argument to the function with the path to the cache directory, and provide a reasonable default.

lookout/style/typos/general.py
irinakhismatullina: It's not, at least with this package.

lookout/style/typos/general.py
irinakhismatullina: @vmarkovtsev Btw, I use pretrained vectors from [here](https://github.com/facebookresearch/fastText/blob/master/docs/pretrained-vectors.md), so I should download them every time to train new fasttext model? It's quite heavy (6Gb). We definitely should provide this functionaly, but I don't think it's good to do it every time in `prepare_data`, it's better to use our pretrained ones in most cases.

vmarkovtsev:
There is also https://developers.google.com/drive/api/v3/manage-downloads which accepts a file object which you can easily wrap to track progress, too.

irinakhismatullina:
@vmarkovtsev I've tried several options before, the best you can get is to show the loaded portion size (without the percentage). Several libraries that I've tried (urllib or requests) couldn't get the total size of the file (and with non-drive files they could and I managed to get a nice progress bar with them).

Probably smth in Drive API will do the trick, but this will require adding several extra libraries to dependencies, does it worth it?

lookout/style/typos/datasets_preparation.py
```diff
@@ -24,6 +26,19 @@
...
 
+class _DownloadProgressBar(tqdm):
+    def update_to(self, b: int = 1, bsize: int = 1, tsize: Optional[int] = None) -> None:
+        if tsize is not None:
+            self.total = tsize
+        self.update(b * bsize - self.n)
+
+
+def _download_url(url: str, output_path: str) -> None:
+    with _DownloadProgressBar(unit="B", unit_scale=True,
```
vmarkovtsev: Can you please switch the units to MB?

Issue #521, opened 2019-01-10
Refactor VirtualNode with annotations


vmarkovtsev:
Inspired by UIMA. We never remove annotations, only add.

...

zurk:
I think that this refactoring issue is an important one to do. We do not have time to implement UIMA approach for Python, so, I think that the easiest solution will be to create a new class `AnnotatedVirtualNode`:
```python
class AnnotatedVirtualNode:
    def __init__():
        self.node = None  # type: VirtualNode
        self.y = None  # type: Sequence[int]
        self.y_pred = None  # type: Sequence[int]
        self.parent = None  # AnnotatedVirtualNode
        # etc
```
And put here all annotations we may need. After that, we agreed that `VirtualNode` is immutable and that is. 

It is a simple approach and maybe I missed some important details why it does not suit well, what do you think @m09?

m09:
This is mostly what we currently have. Some problems include:

- the number and location of attributes is not always the same (basic formatting classes are more numerous than merged formatting classes. Merged formatting classes are more numerous than filtered merged formatting classes, etc)
- this relies on us not modifying values, it's not enforced even slightly by design
- type annotations are not used to their full power (`Optional[…]` everywhere)
- single annotations would be coupled with the full type system (ie all possible annotations).

Overall it's very far from a good annotation pattern. But indeed it might still be good to fix our current VirtualNode in this way (ie remove the cases that are not add-only in the codebase)

zurk:
ok, I knew that I just did not see the problem in all colors. :)
`AnnotatedVirtualNode` v2:
```python
class Annotation:
    def __init__(self):
        self._annotations = {}

    def __getitem__(self, item: str):
        return self._annotations[item]

    def __setitem__(self, key: str, value):
        # We can also add a list of known annotations with their types.
        # So we can check unexpected annotations and if it is known its type.
        if key in self._annotations:
            raise KeyError(("Annotation %s exists with value %s. It is not possible to overwrite "
                           "annotation") % (key, repr(self[key])))
        self._annotations[key] = value

    # I think it will be convenient to have such properties.
    @property
    def y(self):
        try:
            return self._annotations["y"]
        except KeyError:
            raise KeyError("There is no `y` annotation vnode %s")
        
    # etc


class AnnotatedVirtualNode:
    def __init__(self):
        # plus read-only properties
        self._node = None  # type: VirtualNode
        self._annotations = Annotation()
```
[...]

PR #654, opened 2019-02-28
Add annotations

zurk:
Note that all work in progress. However, I created this PR to discuss the proposed solution, find all possible drawbacks and fix them on the early stage. Please note, that the code is raw in minor details. Let's discuss high-level questions about implemented architecture/interface/API decisions.

Let me remind you the main idea of annotations:
**Separate data (code) and all it annotations (like UAST nodes, Virtual Token, `y`, file path, etc) handling.**

This PR contains:
* Annotation tooling. 
    * Will be moved to `lookout-sdk-ml` when I create ready-to-merge PR. For now, I do not want to split the code because it is harder to review. 
    * I cut all possible corners, did not implement a lot of useful functions, did not have good tests, etc. 
* Backward capability layer is added. `to_vnode()` function.
* `FeatureExtractor._parse_vnodes()`, 
`FeatureExtractor._parse_file()`,
`VirtualNode.from_node()` 
functions are rewritten using Annotations to give a feeling of how it will look like. To make everything works I use  `to_vnode()` Backward capability function.

I also checked that I get exactly the same vnodes that we have in current master. 

Related PR: https://github.com/src-d/style-analyzer/issues/521

cc @src-d/machine-learning and specially @m09 PTAL
Any comments and feedback appreciated.


zurk:
The biggest problem I am aware of is dealing with null-intervals `[x, x)` aka NOOPs. I write `[` and `)` because it is semi-intervals like in python slicing. 

There is a lot of cases which require different behavior from null-intervals:
1. Null-intervals should be not covered by the UAST Node annotation next to it. For example, we have `var t = 0;` and there is a NOOP `[9, 9)` between `0` and `;`. Also, there is an UASTNode annotation for `0` `[8, 9)`. As soon as we do not have UAST annotation for NOOP `[9, 9)` we should consider these intervals as disjoint ones.

2. NOOP in the beginning and end of the file. Should be considered as intersecting with file annotations like PATH and LANGUAGE. The same for LINE annotation (annotates the file line number).

3. There are two identical NOOPs between the files: one from the beginning of the first file and second from the end of the next file. However, it is easy to fix. We should add document separator symbol and do not concatenate files directly to each other. So we have the end NOOP at `[len(file1), len(file1))` and the start NOOP at `[len(file1)+1, len(file1))+1`.

    3.1. That means I should move `_check_interval_crossing` from `AnnotatedData` to `Annotated` class and make it dependent from annotation type.

    3.2. There is also another option to store different files in different `AnnotatedData` classes, so `RawData` takes only one document. I feel that this is a bad solution. Once we want to add cross-file annotations (like containing directory) it is better to have all files in one place.

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,265 @@
...
+    All special utilities to work with annotations should be implemented in this class
+    List of methods that should be implemented can be found here:
+    https://uima.apache.org/d/uimafit-current/api/org/apache/uima/fit/util/JCasUtil.html
+    """
+
+    def __init__(self, raw_data: RawData):
+        """Init."""
+        self._raw_data = raw_data
+        # Interval trees should be used for _range_to_annotations later.
+        self._range_to_annotations = SortedDict()  # type: SortedDict[(int, int), Dict[str, Annotation]]  # noqa E501
```
vmarkovtsev: Why do we need to sort?

m09:
Thanks for the first draft @zurk :+1:

Some feedback I can give from UIMA experience:

- it's better to KISS and have a single file per AnnotedData. Then if/when we need cross-files info sharing, we can use external objects.
- we should store per annotation type and not per offset: the goal is to use as few annotation types as possible in a given piece of code, the architecture of the code should reflect that
- for the point about UAST not being applicable to noops: we can query related annotations per covering/covered/equal spans when we want / when it's appropriate, but we can also just point to another annotation in an annotation (this way we specify explicitely which annotation is related):
```python
class TokenAnnotation(Annotation):
    UASTNode: Optional[UASTNodeAnnotation] = None
```
    then when we iterate `TokenAnnotation`s we can retrieve the UAST node this way.
- better not to have `ValuedAnnotation` and specify a meaningful name for each kind of annotation instead of `value`
- better not to have names but just use the classes: names make the code less typesafe
- an annotation containing the complete uast should not be necessary if we have proper `UASTNodeAnnotation`s: only a point to the root `UASTNodeAnnotation` should be enough (in one of the document wide annotations for example).

lookout/style/format/annotations/annotated_data.py
zurk: Because it is possible to add annotations in any order. And I want to have a fast way to iterate through annotations.

zurk:
First feedback addressed, the code becomes more verbose but I and Hugo agreed that it is better to be more explicit then short. 
@m09 hare a day off tomorrow and I think to rewrite couple more functions, add a test, add doc, etc.
So make PR closer to be merged.

lookout/style/format/annotations/annotated_data.py
vmarkovtsev: Yep, but why do we care about this order? We can iterate in any order and it should not matter, should it

zurk: The most common use-case for (Ordered) List of VirtualNodes we have right now is exactly iterating through it in the correct order. We are going to replace this List with Annotations and I want to be able to iterate fast trough it, so I use ordered Dict for that. I do not need to sort to iterate through it. 

It also allows us to search for overlapping/covered/equal annotations by log time. This is also a common usecase.

zurk:
I was busy with https://github.com/src-d/style-analyzer/issues/666 so I do all my plans after a review.

lookout/style/format/annotations/annotated_data.py
vmarkovtsev: Thanks for explaining

vmarkovtsev: May be worth putting this as a comment in the code

m09:
This starts to look ready :+1:. It should be better to have parent attribute in UASTAnnotation and remove the UASTParentAnnotation though, it will make the code much simpler!

zurk:
> It should be better to have parent attribute in UASTAnnotation and remove the UASTParentAnnotation though

I agree. However, when I start implementing it I see that it is not so easy. The problem is that we add the UAST and parents annotation in different places. It looks like we can move parents annotation code to uast annotation, but we have a `uast_fixers` that can change some nodes (which is not happening right now, but it is possible). So only second option left: move UAST annotation code to parents annotation. But it breaks `AnnotatedData.from_file` function because it requires a UAST and it will be really unobvious if you do not add it inside this function.

I hope once v3 python client will be introduced we will have parent feature included into UAST itself and remove `UASTParentAnnotation` at all. So let's keep it as is for now. I do not see big drawbacks for such solution.

zurk:
I think we can consider this PR as done and ready to merge.
I move out all untested and raw changes to the next PRs. 

I am also doing a regression test for this part one more time to be sure that I have exactly the same VNodes in the end. Here is the commit to run the test (I run `make quality-report` and check the logs).
So far no errors (13/20).

m09:
> So only second option left: move UAST annotation code to parents annotation. But it breaks AnnotatedData.from_file function because it requires a UAST and it will be really unobvious if you do not add it inside this function.

This option seems perfectly fine to me. I feel we should do it to avoid complication of the rest of the codebase.

zurk:
Ok, I think I know the answer. 
Actually, we should not allow changing Node in the `uast_fixers`. It is too much. 
Because it actually breaks the UAST itself because you did not replace the old node with a new node.
That means that the old node is still available when you traverse UAST (and you miss a new one).

My plan is:
1. Do not return Node from  `uast_fixers`.
2. Extract parents calculation from `_parse_file` to separate function.
3. Add parents annotation to `AnnotatedData.from_file`

Logic was not broken, code improved and everyone seems to be happy. :)
WDYT @m09?

m09:
yes it seems like great solution: then the code will be easy to write everywhere else :+1:

zurk:
Done.

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+    """Raises by AnnotatedData.find_intersect() if there is no intersection."""
+
+
+class AnnotationsSlice(dict):
+    """
+    Annotations collection for a specific range.
+    """
+
+    def __init__(self, start, stop, *args, **kwargs):
+        """Init."""
```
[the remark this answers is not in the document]
vmarkovtsev: \+ Need to describe the arguments. **Especially** the fact that `stop` is not included.

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
+"""Annotations for style-analyzer."""
+from typing import FrozenSet, Optional, Tuple
+
+from lookout.style.format.utils import get_uast_parents
+
+
+class Annotation:
+    """Base class for annotation."""
```
vmarkovtsev:
```suggestion
    """Base class of all annotations."""
```

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, start: int, stop: int):
+        """
+        Initialization.
```
vmarkovtsev:
```suggestion
        Initialize a new instance blah-blah
```

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+        :param start: Annotation start.
+        :param stop: Annotation end.
+        """
+        self._range = (start, stop)
```
vmarkovtsev: We should check for crazy values here:

- negative
- not in order
- huge

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        :return: Annotated data slice with requested annotations.
+        """
+        for annotations in self.iter_annotations(types, start_offset, stop_offset):
+            yield self[annotations.range], annotations
+
+    def find_intersect(self, name: Type[Annotation], start: int, stop: int) -> Annotation:
+        """
+        Find an annotation of given type that intersects the interval [start, stop).
+
+        raises NoIntersection exception if there is no such annotation.
```
vmarkovtsev: Use `:raise:` or `:error:` docstring, don't remember which exactly

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        :param name: Annotation type.
+        :param start: start of interval.
+        :param stop: end of interval.
+        :return: requested Annotation.
+        """
+        try:
+            annotation_layer = self._type_to_annotations[name]
+        except KeyError:
+            raise NoIntersection("There is no annotation layer %s" % name)
+        search_start = max(0, annotation_layer.bisect_left((start, start)) - 1)
```
vmarkovtsev: Should check crazy arguments

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        search_stop = annotation_layer.bisect_right((stop, stop))
+        for range in annotation_layer.islice(search_start, search_stop):
+            if self._check_interval_crossing(start, stop, *range):
+                # assuming that there is only one such annotation
+                return annotation_layer[range]
+        raise NoIntersection("There is no annotation %s from %d to %d" % (name, start, stop))
+
+    @classmethod
+    def _check_interval_crossing(cls, start1: int, stop1: int, start2: int, stop2: int) -> bool:
+        # TODO(zurk): explain logic with [x, x) intervals.
```
vmarkovtsev: Hehe, yes please! 😄

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __str__(self):
+        return "%s[%d, %d)" % (self.name, self.start, self.stop)
+
+
+class LineAnnotation(Annotation):
+    """Line number annotation."""
+
+    def __init__(self, start: int, stop: int, number: int):
+        """Init."""
```
vmarkovtsev: Initialize a new instance ...

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __repr__(self):
+        return self.__str__()
+
+    def __str__(self):
+        return "%s[%d, %d)" % (self.name, self.start, self.stop)
+
+
+class LineAnnotation(Annotation):
+    """Line number annotation."""
```
vmarkovtsev: Annotate with line numbers.

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, start: int, stop: int, number: int):
+        """Init."""
+        super().__init__(start, stop)
+        self._number = number
+
+    number = property(lambda self: self._number)
```
vmarkovtsev: How about `line`? IMHO fits better.

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        return "%s[%d, %d)" % (self.name, self.start, self.stop)
+
+
+class LineAnnotation(Annotation):
+    """Line number annotation."""
+
+    def __init__(self, start: int, stop: int, number: int):
+        """Init."""
+        super().__init__(start, stop)
+        self._number = number
```
vmarkovtsev: check for a crazy line number; we should mention in the docstring that we are either 0-based or 1-based.

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    number = property(lambda self: self._number)
+
+
+class UASTNodeAnnotation(Annotation):
+    """UAST Node annotation."""
+
+    def __init__(self, start: int, stop: int, node: "bblfsh.Node"):
+        """Init."""
```
vmarkovtsev: ...

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        super().__init__(start, stop)
+        self._node = node
```
vmarkovtsev: Shall we accept `None`?

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+    node = property(lambda self: self._node)
+
+    @staticmethod
+    def from_node(node: "bblfsh.Node") -> "UASTNodeAnnotation":
+        """Create the annotation from bblfsh node."""
+        return UASTNodeAnnotation(node.start_position.offset, node.end_position.offset, node)
+
+
+class UASTAnnotation(UASTNodeAnnotation):
+    """Full UAST of the file annotation."""
```
vmarkovtsev:
```suggestion
    """Whole file UAST annotation."""
```

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+    parents = property(lambda self: self._parents)
+
+
+class TokenAnnotation(Annotation):
+    """Annotation for virtual token in the сode."""
+
+    def __init__(self, start: int, stop: int,
+                 uast_annotation: Optional[UASTNodeAnnotation] = None):
+        """
+        Initialization.
```
vmarkovtsev: ...

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        super().__init__(start, stop)
+        self._uast_annotation = uast_annotation
+
+    uast_annotation = property(lambda self: self._uast_annotation)
+
+    @property
+    def node(self) -> "bblfsh.Node":
+        """
+        Get UAST Node from related UASTNodeAnnotation.
```
vmarkovtsev:
```suggestion
        Get the UAST Node belonging to the underlying `UASTNodeAnnotation`.
```

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+        :return: related bblfsh UAST Node. None if there is no related annotation.
+        """
+        return self._uast_annotation.node if self._uast_annotation else None
```
vmarkovtsev: So we literally annotate with another annotation. Why not carry the node directly?

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    @property
+    def has_node(self) -> bool:
+        """Check if token annotation has related UAST node annotation."""
+        return self._uast_annotation is not None
+
+
+class AtomicTokenAnnotation(TokenAnnotation):
+    """Annotation for undividable tokens generated by FeatureExtractor._classify_vnodes()."""
```
vmarkovtsev: "undividable" means `cannot be divided without leaving a remainder`. You meant "indivisible"

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def to_token_annotation(self) -> TokenAnnotation:
+        """Convert to token annotation."""
```
vmarkovtsev:
```suggestion
        """Convert to a token annotation."""
```

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        return TokenAnnotation(self.start, self.stop, self._uast_annotation)
+
+
+class RawTokenAnnotation(TokenAnnotation):
+    """Annotation for raw tokens generated by FeatureExtractor._parse_file()."""
+
+    def to_atomic_token_annotation(self) -> AtomicTokenAnnotation:
+        """Convert to atomic token annotation."""
```
vmarkovtsev:
```suggestion
        """Convert to an atomic token annotation."""
```

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        return AtomicTokenAnnotation(self.start, self.stop, self._uast_annotation)
+
+
+class LanguageAnnotation(Annotation):
+    """Language of the file annotation."""
+
+    def __init__(self, start: int, stop: int, language: str):
+        """Init."""
```
vmarkovtsev: ...

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        super().__init__(start, stop)
+        self._language = language
+
+    language = property(lambda self: self._language)
+
+
+class PathAnnotation(Annotation):
+    """File language annotation."""
```
vmarkovtsev: language?

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, start: int, stop: int, path: str):
+        """Init."""
```
vmarkovtsev: ...

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        super().__init__(start, stop)
+        self._path = path
+
+    path = property(lambda self: self._path)
+
+
+class TargetAnnotation(Annotation):
+    """Target for model prediction annotation."""
```
vmarkovtsev:
```suggestion
    """Can(should) be predicted by a model."""
```

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, start: int, stop: int, target: Tuple[int, ...]):
+        """Init."""
```
vmarkovtsev: ...

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        super().__init__(start, stop)
+        self._target = target
+
+    target = property(lambda self: self._target)
+
+
+class AtomicTargetAnnotation(TargetAnnotation):
```
vmarkovtsev: Let's use "class" instead of "atomic target" and "label" instead of "target" everywhere. This corresponds to the terms in our paper.

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+    """Annotation for atomic format symbols like quote, tab, space, etc."""
+
+    def to_target_annotation(self) -> TargetAnnotation:
+        """Convert annotation to TargetAnnotation."""
+        return TargetAnnotation(self.start, self.stop, self._target)
+
+
+class AccumulatedIndentationAnnotation(Annotation):
+    """Annotates accumulated indentation Token."""
```
vmarkovtsev:
```suggestion
    """Annotates an "accumulated indentation" token."""
```

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class LinesToCheckAnnotation(Annotation):
+    """Annotation to store set of lines where the style should be checked."""
+
+    def __init__(self, start: int, stop: int, lines: FrozenSet[int]):
+        """Init."""
```
vmarkovtsev: ...

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        """Convert annotation to TargetAnnotation."""
+        return TargetAnnotation(self.start, self.stop, self._target)
+
+
+class AccumulatedIndentationAnnotation(Annotation):
+    """Annotates accumulated indentation Token."""
+
+
+class LinesToCheckAnnotation(Annotation):
+    """Annotation to store set of lines where the style should be checked."""
```
vmarkovtsev:
```suggestion
    """Annotation to store a set of lines where the style shall be checked."""
```

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+        self._lines = lines
+
+    lines = property(lambda self: self._lines)
+
+
+class TokenParentAnnotation(Annotation):
+    """Annotation to store token UAST parents."""
+
+    def __init__(self, start: int, stop: int, parent: "bblfsh.Node"):
+        """Init."""
```
vmarkovtsev: ...

lookout/style/format/annotations/annotations.py
```diff
@@ -0,0 +1,196 @@
...
+    def __init__(self, start: int, stop: int, lines: FrozenSet[int]):
+        """Init."""
+        super().__init__(start, stop)
+        self._lines = lines
+
+    lines = property(lambda self: self._lines)
+
+
+class TokenParentAnnotation(Annotation):
+    """Annotation to store token UAST parents."""
```
vmarkovtsev:
```suggestion
    """Annotation to store the UAST parent node of a token."""
```

lookout/style/format/feature_extractor.py
```diff
@@ -307,22 +310,23 @@ def _parse_vnodes(self, files: Iterable[File], lines: Optional[List[List[int]]]
         parsed_files = []
         index_labels = not self.labels_to_class_sequences
         for i, file in enumerate(files):
-            contents = file.content.decode("utf-8", "replace")
+            path = file.path
             uast = file.uast
+            file = AnnotatedData.from_file(file)
             try:
-                file_vnodes, file_parents = self._parse_file(contents, uast, file.path)
+                # Should be self._parse_file(content, file)
```
vmarkovtsev: wdym "should be"? TODO or FIXME?

lookout/style/format/feature_extractor.py
```diff
@@ -307,22 +310,23 @@ def _parse_vnodes(self, files: Iterable[File], lines: Optional[List[List[int]]]
...
+                file_vnodes, file_parents = self._parse_file(file)
             except AssertionError as e:
-                self._log.warning("could not parse file %s with error '%s', skipping",
-                                  file.path, e)
+                self._log.warning("could not parse file %s with error '%s', skipping", path, e)
```
vmarkovtsev:
```suggestion
                self._log.warning("could not parse %s: error '%s', skipping", path, e)
```

lookout/style/format/feature_extractor.py
```diff
@@ -689,21 +693,19 @@ def _keep_sibling(self, sibling: VirtualNode, vnode: VirtualNode, include_labele
         quote_classes = set([CLASS_INDEX[CLS_DOUBLE_QUOTE], CLASS_INDEX[CLS_SINGLE_QUOTE]])
         return not (quote_classes & set(vnode.y) and quote_classes & set(sibling.y))
 
-    def _parse_file(self, contents: str, root: bblfsh.Node, path: str) -> \
-            Tuple[List[VirtualNode], Dict[int, bblfsh.Node]]:
+    def _parse_file(self, file: AnnotatedData) -> Tuple[List[VirtualNode], Dict[int, bblfsh.Node]]:
         """
-        Parse a file into a sequence of `VirtuaNode`-s and a mapping from VirtualNode to parent.
+        Annotate a file with a RawTokenAnnotation and build a mapping from UAST Node to parent.
```
vmarkovtsev:
```suggestion
        Annotate a file with `RawTokenAnnotation`-s and build a mapping from UAST nodes to their parents.
```

lookout/style/format/feature_extractor.py
```diff
@@ -689,21 +693,19 @@ def _keep_sibling(self, sibling: VirtualNode, vnode: VirtualNode, include_labele
...
+        `RawTokenAnnotation`-s and builds the parents mapping. `parents` map from `id(node)` to
+        its parent `bblfsh.Node`.
 
-        :param contents: source file text
-        :param root: UAST root node
-        :param path: path to the file, used for debugging
+        :param file: Source code annotated with path and uast.
         :return: list of `VirtualNode`-s and the parents.
         """
+        # TODO(zurk): rename this function when refactoring is done.
```
vmarkovtsev:
```suggestion
        # TODO(zurk): rename this function when the refactoring is finished.
```

lookout/style/format/feature_extractor.py
```diff
@@ -689,21 +693,19 @@ def _keep_sibling(self, sibling: VirtualNode, vnode: VirtualNode, include_labele
...
-        source text bit-to-bit after `"".join(n.value for n in nodes)`. `parents` map from
-        `id(node)` to its parent `bblfsh.Node`.
+        Given the source text and the corresponding UAST this function covers all code with a
+        `RawTokenAnnotation`-s and builds the parents mapping. `parents` map from `id(node)` to
+        its parent `bblfsh.Node`.
 
-        :param contents: source file text
-        :param root: UAST root node
-        :param path: path to the file, used for debugging
+        :param file: Source code annotated with path and uast.
```
vmarkovtsev:
```suggestion
        :param file: Source code annotated with `PathAnnotation`-s and `???`-s.
```

lookout/style/format/feature_extractor.py
```diff
@@ -742,33 +741,31 @@ def _parse_file(self, contents: str, root: bblfsh.Node, path: str) -> \
...
                     "missed some imaginary tokens: \"%s\"" % diff
             if node is sentinel:
                 break
-            result.extend(VirtualNode.from_node(node, contents, path, self.token_unwrappers))
+            uast_node_annot = list(VirtualNode.from_node(node, contents, self.token_unwrappers))
+            file.update(uast_node_annot)
             pos = node.end_position.offset
-        return result, parents
+
+        # backward
```
vmarkovtsev: backward?

lookout/style/format/feature_extractor.py
```diff
@@ -806,3 +803,42 @@ def _fill_vnode_parents(self, file_parents: Mapping[int, bblfsh.Node],
             if parent is None:
                 parent = uast
             vnode_parents[id(vn)] = parent
+
+
+def raw_file_to_vnodes_and_parents(file: AnnotatedData) -> Tuple[List["VirtualNode"],
+                                                                 Dict[int, bblfsh.Node]]:
+    """
+    Convert AnnotatedData class to Sequence of vnodes.
```
vmarkovtsev:
```suggestion
    Convert AnnotatedData class to a sequence of vnodes.
```

lookout/style/format/feature_extractor.py
```diff
@@ -806,3 +803,42 @@ def _fill_vnode_parents(self, file_parents: Mapping[int, bblfsh.Node],
...
+
+    For backward capability.
```
vmarkovtsev:
```suggestion
    For backwards capability.
```

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        self._stop = stop
+
+    start = property(lambda self: self._start)
+
+    stop = property(lambda self: self._stop)
+
+    range = property(lambda self: self._range)
+
+
+class AnnotatedData:
```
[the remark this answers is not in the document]
zurk: if @m09 is ok with this name.

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        Get part of content for a specific index range.
+
+        :param item: index or index range.
+        :return: Corresponding part of content.
+        """
+        if isinstance(item, tuple):
+            item = slice(*item)
+        if isinstance(item, slice) and item.step is not None:
+            raise KeyError("slice.step is not supported.")
+        return self._content[item]
```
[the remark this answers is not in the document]
zurk: No, we do not. A string is immutable in python, so it is copied automatically

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        """
+        for annotation in annotations:
+            self.add(annotation)
+
+    def get(self, annotation_type: Type[Annotation], range: Optional[Tuple[int, int]] = None,
+            ) -> Annotation:
+        """
+        Return a specific annotation for a given range.
+
+        Looking for exact match only. If range is None it returns annotations that cover all
```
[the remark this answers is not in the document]
zurk: both depending if range (span) is provided. I'll clarify this.

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        if start_offset is not None:
+            search_from = self._type_to_annotations[name].bisect_left(
+                (start_offset, start_offset))
+            for value in self._type_to_annotations[name].values()[search_from:]:
+                yield value
+        else:
+            for value in self._type_to_annotations[name].values():
+                yield value
+
+    def iter_annotations(self, types: Sequence[Type[Annotation]],
```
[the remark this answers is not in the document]
zurk: either you want to iterate through only one annotation type or several.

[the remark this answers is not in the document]
zurk: same logic as with `add` and `update` functions.

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+                except NoIntersection:
+                    pass
+            annotations.update({type: same_range_annotations[type] for type in common})
+            yield AnnotationsSlice(*annotation0.range, annotations)
+
+    def iter_items(self, types: Sequence[Type[Annotation]], start_offset: Optional[int] = None,
+                   stop_offset: Optional[int] = None,
+                   ) -> Iterator[Tuple[str, AnnotationsSlice]]:
+        """
+        Iterate through annotations with specified type.
```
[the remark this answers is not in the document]
zurk: yes. I want to make this tool nice to use. 
I have a standard python `dict` interface in mind.
that why I have `update` function.

this one is a `dict.items()`

lookout/style/format/annotations/annotations.py
zurk: No, we just do not have an annotation in this case.

lookout/style/format/annotations/annotations.py
zurk: Well, there is an underlying discussion with @m09 https://github.com/src-d/style-analyzer/pull/654#issuecomment-469636019 
@m09, can you give me help to answer this?

lookout/style/format/annotations/annotations.py
zurk: Javascript, Python, Go

lookout/style/format/annotations/annotated_data.py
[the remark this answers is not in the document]
m09: :+1:

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        """
+        Add annotation. One type annotations can not overlap with each other.
+        """
+        annotation_id = type(annotation)
+        if annotation.start == 0 and annotation.stop == len(self):
+            if annotation_id in self._global_annotations:
+                raise ValueError("Global annotation %s already exists" % annotation)
+            self._global_annotations[annotation_id] = annotation
+        else:
+            # TODO(zurk): Add a check that there is no overlapping annotations of one type.
```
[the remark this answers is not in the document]
m09: I think it's beneficial to have overlapping annotations (for example it's required for UASTAnnotation)

lookout/style/format/feature_extractor.py
zurk: yes, I call a function to convert a new format to old vnodes.

I'll elaborate the comment

lookout/style/format/annotations/annotations.py
m09: For me it's better to link to the annotation (not the bblfsh node). The goal of UASTNodeAnnotations being to enhance bblfsh.Node for our use case (here with parents for example, and potentially role/internalrole mapping, etc).

lookout/style/format/feature_extractor.py
zurk: I mean when I finish refactoring. I address this comment in the next PR
Here I decided to keep `_parse_file` signature the same because I do not want to change tests now.

zurk:
As soon as there is a lot of conflicts I am going to close this one and open new PR.
All feedback from the review will be addressed.

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        self._content = content
+
+        # Dictionary to store annotations for all file (aka `global` annotations)
+        self._global_annotations = {}  # type: Dict[Type[Annotation], Annotation]
+
+        # _range_to_annotations dict is created for optimization purpose only.
+        # The most common use-case we have in style-analyzer is iterating through Token annotations
+        # in the sorted order. To iterate fast ordered Dict is used.
+        self._range_to_annotations = SortedDict()  # type: SortedDict[(int, int), Dict[Type[Annotation], Annotation]]  # noqa E501
+        self._type_to_annotations = {}  # type: Dict[Type[Annotation], SortedDict[(int, int), Annotation]]  # noqa E501
```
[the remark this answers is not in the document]
zurk: I add a big comment to make everything clear.

lookout/style/format/annotations/annotated_data.py
```diff
@@ -0,0 +1,230 @@
...
+        :param item: index or index range.
+        :return: Corresponding part of content.
+        """
+        if isinstance(item, tuple):
+            item = slice(*item)
+        if isinstance(item, slice) and item.step is not None:
+            raise KeyError("slice.step is not supported.")
+        return self._content[item]
+
+    def count(self, annotation_type: Type[Annotation]):
```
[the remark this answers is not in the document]
zurk: there is nothing else to count. Can I keep it?

lookout/style/format/annotations/annotated_data.py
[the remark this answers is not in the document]
zurk: I think we should add support for the overlapping annotations but it is hard to do right now.

Issue #658, opened 2019-03-05
Refactor QualityReport 


zurk:
We should finish `QualityReport` and `GeneralReport` refactoring so they become easily extendable and without code duplication, we have right now: https://github.com/src-d/style-analyzer/blob/master/lookout/style/format/benchmarks/quality_report.py#L370-L378

PR #665, opened 2019-03-06
Add GeneralReports.generate_reports() function

zurk:
First PR for the issue: https://github.com/src-d/style-analyzer/issues/658
PR priority: low. 

There are a lot of things to improve in our report generation system. When my main issue was blocked I refactor 1 thing: possibility to generate as many reports as you want with `GeneralReport` class. Before it was hardcoded to train, test and model reports. 

Logic is unchanged. Just one minor thing: Model report generated only in `aggregate` mode.
`make report-quality` logic completely the same.


lookout/style/format/benchmarks/quality_report.py
```diff
@@ -119,25 +99,23 @@ def _convert_files_to_xy(self, parsed_files):
...
 
-    def capture_report(func, name):
+    reports = {}
+
+    def capture_reports(func):
         @functools.wraps(func)
-        def wrapped_capture_quality_report(*args, **kwargs):
-            if getattr(report, name) is not None:
-                raise RuntimeError("%s should be called only one time." % func.__name__)
+        def wrapped_capture_quality_reports(*args, **kwargs):
```
vmarkovtsev: One day we should remove this function patching at all

zurk: I also do not like it. 
If we found any way to access comments from lookout-sdk without parsing logs we can easily remove this decorator.

Issue #667, opened 2019-03-06
Optimize the typos tests


vmarkovtsev:
- Reduce the amount of work done
- Mark with `@long-test` those which have to be long-running
- Hack `.travis.yml` to spawn separate jobs with typos tests

vmarkovtsev:
Done

PR #668, opened 2019-03-06
Optimize typos tests

irinakhismatullina:
In accordance with #667.

- Replaced test FastText model with a very small one.
- Loaded to drive a small raw dataset to speed up test for prepare_dataset() with loading file from drive.
- Added `@long_test` decorator to a couple of tests.


irinakhismatullina:
I don't understand why it can't see changed bin file in out-of-docker builds. Any ideas @zurk @vmarkovtsev?

vmarkovtsev:
Still 25 minutes, does not seem to help.

irinakhismatullina:
When I run all typos tests locally, it takes 3 seconds (without installs and setting up env, just tests). Can it be Travis, or too much other builds?

vmarkovtsev:
@irinakhismatullina Please refer to the Travis logs, it reports the time taken by each command.

irinakhismatullina:
@vmarkovtsev Well, here total for all tests (format + typos) is 920s, 15 min. It doesn't show time for smaller chunks, or I couldn't find it. 
Also I still don't understand why it fails out of docker here. Logs tell that the file can't be found, i just replaced it. Everything runs alright locally out of docker and inside the docker in Travis.

zurk:
Add your `test_ft.bin` to
https://github.com/src-d/style-analyzer/blob/651ee1664fb074b0dded87c3e10c6e340f5dd5f7/setup.py#L79

Should help.

irinakhismatullina:
Wow, that's strange, thank you.

vmarkovtsev:
@irinakhismatullina Here is what I see

screenshot 2019-03-07 at 10 37 04

Python 3.5 is doing great, however, 3.6 and 3.7 are very, very slow. You need to go through the log timings and make a report.

screenshot 2019-03-07 at 10 38 32

vmarkovtsev:
> Wow, that's strange, thank you.

The reason is that we test the built package, and you tested the local files. Extra files must be included into the package, otherwise, they are not installed. We had precisely the same problem in the past and that's why we actually test the package now.

zurk:
Something is strange. I think it can be Travis issues. Let's restart are see if we have the same result?

irinakhismatullina:
> The reason is that we test the built package, and you tested the local files. Extra files must be included into the package, otherwise, they are not installed. We had precisely the same problem in the past and that's why we actually test the package now.

Yes, I understand, it's just that I definitely replaced the filename in setup.py, but looks like that change was somehow lost in the final commit. My bad.

zurk:
Ok, @vmarkovtsev it is because of https://github.com/src-d/style-analyzer/blob/master/lookout/style/format/tests/test_quality_report_noisy.py

Since we always retrain the model for a noisy report it takes a huge amount of time in a test also.
I am going to fix it ASAP.  (In Python 3.5 we just skip this test because of difflib)

vmarkovtsev:
OK, this means that the original PR is fine. Merging.

lookout/style/typos/tests/test_preparation.py
```diff
@@ -6,24 +6,26 @@
...
 from lookout.style.typos.utils import Columns, read_frequencies, read_vocabulary
 
 
 TEST_DATA_DIR = pathlib.Path(__file__).parent
-FASTTEXT_DUMP_FILE = str(TEST_DATA_DIR / "id_vecs_10.bin")
-VOCABULARY_FILE = str(TEST_DATA_DIR / "test_frequencies.csv.xz")
 
 
 class PreparationTest(unittest.TestCase):
+    @long_test
```
vmarkovtsev: Actually, we can remove this because you significantly reduced the test size below

lookout/style/typos/tests/test_preparation.py
```diff
@@ -72,11 +74,12 @@ def test_get_datasets(self):
 
 
 class FasttextTest(unittest.TestCase):
+    @long_test
```
vmarkovtsev: Same here - I guess we can always run it

vmarkovtsev:
@irinakhismatullina Tests failed

irinakhismatullina:
I wonder, usually `pandas.read_csv("*.xz")` works fine, but somehow after loading from drive or reading from tempdir it became a problem, that's unexpected.

lookout/style/typos/preparation.py
```diff
@@ -98,7 +98,10 @@ def prepare_data(params: Optional[Mapping[str, Any]] = None) -> pandas.DataFrame
         raw_data_path = os.path.join(params["data_dir"], params["raw_data_filename"])
         _download_url(params["dataset_url"], raw_data_path)
 
-    data = pandas.read_csv(raw_data_path, index_col=0)
+    try:
+        data = pandas.read_csv(raw_data_path, index_col=0)
+    except UnicodeDecodeError:
+        data = pandas.read_csv(raw_data_path, compression="xz", index_col=0)
```
zurk: it should be possible to check file extension like
```python
if raw_data_path.endswith(".xz"):
    ...
else: 
    ...
```
It is a better approach then try-except.

zurk: Now I see your comment: https://github.com/src-d/style-analyzer/pull/668#issuecomment-470580979
Maybe you save file to temp file without `.xz` suffix?

irinakhismatullina: Thanks!

vmarkovtsev: So pandas can read xz archives, nice

irinakhismatullina: > Maybe you save the file to temporary one without `.xz` suffix?

Oh, yes, it must be it, thank you again!

PR #671, opened 2019-03-07
Optimize style-analyzer tests 

zurk:
Test for noisy report mostly.

### DIFF
--- a/lookout/style/format/benchmarks/quality_report_noisy.py
+++ b/lookout/style/format/benchmarks/quality_report_noisy.py
@@ -8,7 +8,7 @@ import os
 import subprocess
 import sys
 import tempfile
-from typing import Iterable, List, Mapping, NamedTuple, Optional, Set, Tuple
+from typing import Iterable, List, Mapping, NamedTuple, Optional, Set, Tuple, Union
 
 from bblfsh import BblfshClient
 import jinja2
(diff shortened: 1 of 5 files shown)


zurk:
@vmarkovtsev do you want to merge style and test checks back? It also improves our Travis speed by 3-5 mins.

zurk:
:muscle: 
image

zurk:
It was accidentally closed.

zurk:
Long tests check: https://travis-ci.org/zurk/style-analyzer/builds/503054071

warenlg:
You could also decrease `n_iter` in the default test config to win even more time, no ?

zurk:
Yes, it is true, but it is already 6. And actually we have 10 evaluations in random points, so setting less than 10 is the same.

vmarkovtsev:
Which style and test checks? 🤔

zurk:
We have two stages in Travis. The first one in style check (`flake8`) the second one is testing. One goes after another sequentially.  As soon as we have more slots in Travis we can do them in parallel as before to speed up Travis running time from ~19 mins to ~15 mins.

vmarkovtsev:
Now I see what you mean, thanks. It saved much resources in the past, so I don't feel that we should revert this... It has an extra nice side effect that you have a fair chance to stop the Travis build before it occupies many slots.

zurk:
Good point. :+1:

PR #681, opened 2019-03-08
Refactor generator

irinakhismatullina:
- Add multiprocessing management features
- Small refactoring of parameters handling
- Fix problems with modelforge binding

### DIFF
--- a/lookout/style/typos/corrector.py
+++ b/lookout/style/typos/corrector.py
@@ -39,6 +39,8 @@ class TyposCorrector(Model):
                            "alpha": 1,
                            "eval_metric": ["error"],
                            "nthread": 0}
+    DEFAULT_START_POOL_SIZE = 64
+    DEFAULT_CHUNKSIZE = 256
 
     def __init__(self, **kwargs):
         """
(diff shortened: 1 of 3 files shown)


lookout/style/typos/corrector.py
```diff
@@ -23,13 +23,13 @@ class TyposCorrector(Model):
...
     DESCRIPTION = "Model that suggests fixes to correct typos."
     LICENSE = DEFAULT_LICENSE
-    DEFAULT_RADIUS = 3
-    DEFAULT_MAX_DISTANCE = 2
-    DEFAULT_NEIGHBORS_NUMBER = 0
-    DEFAULT_EDIT_CANDIDATES = 20
-    DEFAULT_TRAIN_ROUNDS = 4000
-    DEFAULT_EARLY_STOPPING = 200
-    DEFAULT_BOOST_PARAM = {"max_depth": 6,
+    default_radius = 3
```
vmarkovtsev: These should stay capital case, because this is how we mark the constants.

irinakhismatullina: Ok, I wanted to move them to dictionaries (configs) in the next PRs, should I name the default dictionary in capital too? I only started to do lowercase after I saw that in several places in format analyzer, but probably it is changed already.

PR #684, opened 2019-03-11
Add train from scratch function

irinakhismatullina:
Train from scratch function, just wraps up all preparation functions.

I am not shure of the best way to handle parameters here. All the preparation functions will also have wrappings for CL usage. So I prefer passing dictionaries to specifying all the defaults in several places, but I'm not shure it's a good practice.
Will be glad to any advice on the subject.

### DIFF
--- a/lookout/style/typos/preparation.py
+++ b/lookout/style/typos/preparation.py
@@ -1,4 +1,3 @@
-from copy import deepcopy
 import os
 import pathlib
 import sys
@@ -19,16 +18,6 @@ from lookout.style.typos.utils import Columns, flatten_df_by_column
 
 DATA_DIR = pathlib.Path(__file__).parent / "data"
 
-defaults_for_fasttext = {
-    "size": 100000000,
-    "corrupt": True,
-    "typo_probability": 0.2,
-    "add_typo_probability": 0.005,
-    "fasttext_path": str(DATA_DIR / "emb.bin"),
-    "dim": 8,
-    "bucket": 200000,
-}
-
 defaults_for_preparation = {
     "data_dir": str(DATA_DIR),
     "input_path": str(DATA_DIR / "raw_data.csv"),
@@ -40,6 +29,26 @@ defaults_for_preparation = {
     "raw_data_filename": "raw_data.csv",
     "vocabulary_filename": "vocabulary.csv",
     "frequencies_filename": "frequencies.csv",
+    "prepared_filename": "prepared.csv",
+}
+
+defaults_for_fasttext = {
+    "size": 100000000,
+    "corrupt": True,
+    "typo_probability": 0.2,
+    "add_typo_probability": 0.005,
+    "fasttext_path": str(DATA_DIR / "emb.bin"),
+    "dim": 8,
+    "bucket": 200000,
+}
+
+defaults_for_datasets = {
+    "train_size": 50000,
+    "test_size": 10000,
+    "typo_probability": 0.5,
+    "add_typo_probability": 0.01,
+    "train_path": str(DATA_DIR / "train.csv"),
+    "test_path": str(DATA_DIR / "test.csv"),
 }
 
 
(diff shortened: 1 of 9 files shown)


lookout/style/typos/preparation.py
```diff
@@ -40,6 +29,26 @@
     "raw_data_filename": "raw_data.csv",
     "vocabulary_filename": "vocabulary.csv",
     "frequencies_filename": "frequencies.csv",
+    "prepared_filename": "prepared.csv",
+}
+
+defaults_for_fasttext = {
+    "size": 100000000,
```
vmarkovtsev: Would be nice to have a comment here what does this huge size mean

lookout/style/typos/preparation.py
```diff
@@ -40,6 +29,26 @@
...
+    "corrupt": True,
+    "typo_probability": 0.2,
+    "add_typo_probability": 0.005,
+    "fasttext_path": str(DATA_DIR / "emb.bin"),
```
vmarkovtsev: Better to use long names: embeddings.bin

And btw what is `bin`? If it is the numpy format, we should use `npy`, etc.

lookout/style/typos/preparation.py
```diff
@@ -84,14 +93,14 @@ def prepare_data(params: Optional[Mapping[str, Any]] = None) -> pandas.DataFrame
                                      frequencies for all present tokens will be saved.
                    raw_data_filename: Name of .csv file in data_dir to put raw dataset in case of \
                                       loading from drive.
-                   vocabulary_path: Name of .csv file in data_dir to save vocabulary to.
-                   frequencies_path: Name of .csv file in data_dir to save frequencies to.
+                   vocabulary_filename: Name of .csv file in data_dir to save vocabulary to.
```
vmarkovtsev:
```suggestion
                   vocabulary_filename: Name of the .csv file in `data_dir` to save vocabulary to.
```

(the same below)

lookout/style/typos/preparation.py
```diff
@@ -200,9 +183,53 @@ def train_fasttext(data: pandas.DataFrame, params: Optional[Mapping[str, Any]] =
...
+        test.to_csv(params["test_path"])
+    if params["train_path"] is not None:
+        train.to_csv(params["train_path"])
+    return train, test
+
+
 def train_and_evaluate(train_data: pandas.DataFrame, test_data: pandas.DataFrame,
                        vocabulary_path: str, frequencies_path: str, fasttext_path: str,
-                       threads_number: int = 8) -> TyposCorrector:
+                       threads_number: int = 16) -> TyposCorrector:
```
vmarkovtsev: You should never hardcode these values. `multiprocessing.cpu_count()`

lookout/style/typos/preparation.py
```diff
@@ -225,3 +252,46 @@ def train_and_evaluate(train_data: pandas.DataFrame, test_data: pandas.DataFrame
...
     model.evaluate(test_data)
     return model
+
+
+def train_from_scratch(prepare_params: Optional[Mapping[str, Any]] = None,
+                       pretrained_fasttext: Optional[str] = None,
+                       fasttext_params: Optional[Mapping[str, Any]] = None,
+                       datasets_params: Optional[Mapping[str, Any]] = None,
+                       save_model_path: Optional[str] = None,
+                       threads_number: int = 16) -> TyposCorrector:
```
vmarkovtsev: same here - `threads_number`

lookout/style/typos/preparation.py
irinakhismatullina: It's fasttext default format for saving full model.

lookout/style/typos/preparation.py
```diff
@@ -200,9 +185,54 @@ def train_fasttext(data: pandas.DataFrame, params: Optional[Mapping[str, Any]] =
...
+                          Columns.Split, Columns.Frequency and Columns.Token.
+    :param params: Parameters for creating train and test datasets, options:
+                   train_size: Train dataset size.
+                   test_size: Test dataset size.
+                   typo_probability: Probability with which a token gets to be corrupted.
+                   add_typo_probability: Probability with which one more corruption happens to a \
+                                         corrupted token.
+                   train_path: Path to the .csv file where to save the train dataset.
+                   test_path: Path to the .csv file where to save the test dataset.
+    :param threads_number: Number of threads for multiprocessing.
```
vmarkovtsev: multiprocessing works wit htprocesses, not with threads. Those two are very different, especially for Python. So do you mean `processes_number` here?

irinakhismatullina: Yes, it's processes, changing it everywhere.

PR #696, opened 2019-03-13
Add expand_vocabulary method to the corrector

irinakhismatullina:
Super small feature: expand the vocabulary in the corrector.

After adding new tokens to the vocabulary through this method, model:
- will be enabled to suggest those new tokens as corrections
- but still will probably suggest to correct them themselves (as being inside the vocabulary is not a green light for the check pass, for we don't have a perfect vocabulary and also there are typos that are correct words of the wrong meaning).

For now i believe that this *never-suggest-corrections-for-me* set must be held in analyzer (will be implementer in the upcoming PR). But we can also try different configurations in the future.

### DIFF
--- a/lookout/style/typos/generation.py
+++ b/lookout/style/typos/generation.py
@@ -47,7 +47,7 @@ class CandidatesGenerator(Model):
         super().__init__(**kwargs)
         self.checker = None
         self.wv = None
-        self.tokens = []
+        self.tokens = set()
         self.frequencies = {}
         self.min_freq = 0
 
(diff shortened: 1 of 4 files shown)


lookout/style/typos/generation.py
```diff
@@ -339,7 +347,7 @@ class DummyModel(Model):
         for key, val in self.checker._words.items():
             wordvals[delstrs_map[key]] = val
         tree["checker"]["_words"] = wordvals
-        tree["tokens"] = merge_strings(self.tokens)
+        tree["tokens"] = merge_strings(list(self.tokens))
```
vmarkovtsev: This should be deterministic:
```suggestion
        tree["tokens"] = merge_strings(sorted(self.tokens))
```

PR #700, opened 2019-03-14
Fix train_from_scratch()

zurk:
This changes are important and without them I was not able to run train_from_scratch()

### DIFF
--- a/lookout/style/typos/preparation.py
+++ b/lookout/style/typos/preparation.py
@@ -105,6 +105,7 @@ def prepare_data(config: Optional[Mapping[str, Any]] = None) -> pandas.DataFrame
         config = {}
     config = merge_dicts(DEFAULT_CONFIG["preparation"], config)
 
+    os.makedirs(config["data_dir"], exist_ok=True)
     raw_data_path = config["input_path"]
     if raw_data_path is None or not os.path.exists(raw_data_path):
         raw_data_path = os.path.join(config["data_dir"],


lookout/style/typos/generation.py
```diff
@@ -388,8 +388,7 @@ def _load_tree(self, tree: dict) -> None:
         self.checker._deletes = deletes
         self.checker._words = {w: self.checker._words[i] for i, w in enumerate(words)}
         vectors = self.wv["vectors"]
-        wv = FastTextKeyedVectors(vectors.shape[1], self.wv["min_n"], self.wv["max_n"],
-                                  self.wv["bucket"], True)
+        wv = FastTextKeyedVectors(vectors.shape[1], self.wv["min_n"], self.wv["max_n"])
```
irinakhismatullina: We shouldn't change this. Most likely this caused problems for you because you have outdated version of gensim. The requirements were updated recently, you should use the version stated there.

lookout/style/typos/preparation.py
```diff
@@ -62,6 +62,7 @@ def update_to(self, b: int = 1, bsize: int = 1, tsize: Optional[int] = None) ->
 
 
 def _download_url(url: str, output_path: str) -> None:
+    os.makedirs(os.path.split(output_path)[0], exist_ok=True)
```
irinakhismatullina: Thank you, this needed to be done, but I suggest moving it to the beginning of `prepare_data()` function, as the same problem (directory doesn't exist) may also happen when the raw dataset is loaded from the local path.

PR #703, opened 2019-03-14
Prepare Typos Analyzer to be evaluated

zurk:
It is the first PR in the next series of PRs that is needed to have a quality report system.
It changes Typos Analyzer to be similar to Format Analyzer.

I extracted one main functions with all logic inside: `generate_typos_fixes` that almost all I need to be able to create a report system.

### DIFF
--- a/lookout/style/typos/analyzer.py
+++ b/lookout/style/typos/analyzer.py
@@ -1,7 +1,7 @@
 """Identifier typos analyzer."""
 from collections import defaultdict
 import logging
-from typing import Any, Dict, List, Mapping, Tuple
+from typing import Any, Dict, Iterable, Iterator, List, Mapping, NamedTuple, Sequence, Tuple
 
 import bblfsh
 from lookout.core.analyzer import Analyzer, AnalyzerModel, DummyAnalyzerModel, ReferencePointer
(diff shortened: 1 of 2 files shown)


lookout/style/typos/analyzer.py
```diff
@@ -1,20 +1,34 @@
...
 from lookout.core.lib import extract_changed_nodes, files_by_language, filter_files, find_new_lines
+from lookout.sdk.service_data_pb2 import Change, File
 import pandas
 from sourced.ml.algorithms import TokenParser, uast2sequence
 
+from lookout.style.format.utils import generate_comment
 from lookout.style.typos.corrector_manager import TyposCorrectorManager
 from lookout.style.typos.utils import Columns, flatten_df_by_column
 
+Candidate = NamedTuple("Candidate", (
```
irinakhismatullina: I think this class should be moved to the `utils.py` in the future and used everywhere through corrector suggestions logic. I can do it after this PR is merged

lookout/style/typos/analyzer.py
```diff
@@ -107,18 +135,31 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
+        # TODO: move this logic to template. See FormatAnalyzer.render_comment_text()
+        corrections_line = ", ".join("%s (%d%%)" % (
+            c.token, int(c.confidence * 100)) for c in typo_fix.candidates)
+        return "Possible typo in \"%s\". Suggestions: %s" % (typo_fix.token, corrections_line)
+
+    @staticmethod
+    def _get_comment_confidence(suggestions: Sequence[Candidate]) -> int:
+        invert_confidence = 1
+        for suggestion in suggestions:
+            invert_confidence *= (1 - suggestion.confidence)
```
irinakhismatullina: Looks like here you multiply probabilities for correcting the same token inside on identifier, while wanted to do it for the whole identifier.
So for example you have `get nomber of noddes`. Model's corrections for this identifier are `nomber: (number, 0.99), (none, 0.10)` and `noddes: (nodes, 0.9), (node, 0.6)`.
And here you return comments `Typo in 'nomber' with confidence (1 - 0.01 * 0.9)` and `Typo in 'noddes' with confidence (1 - 0.1 * 0.4)`. 
Is that exactly what you wanted to do, not multiply probas for different corrections for the whole identifier? Like `Typo in 'get_nomber_of_noddes with confidence (1 - 0.01 * 0.1)`?
If you did want the first way, then why not to take the best candidate probability as the confidence then?

lookout/style/typos/analyzer.py
zurk: Ok, I'll move it to `utils.py`.

lookout/style/typos/analyzer.py
zurk: you are right, it is a probability of having a good suggestion for one token in the identifier. And for now, we have two comments for your example. One per typo. 

So actually, I keep it as it was and we should change it is the future when we are able to reconstruct identifier. Thus, in next PRs. :)

EgorBu: another point - probabilities are not normalized `noddes: (nodes, 0.9), (node, 0.6)`.

zurk: well, we will change it to anything we want when it becomes important. For now, it is ok to have any reasonable number. Anyway we cannot verify it now.

zurk: it is fine they are not normalized. That is why I use black probability theory magic and not just sum it up.

PR #708, opened 2019-03-15
 Use Annotations in _parse_vnodes()

zurk:
Based on https://github.com/src-d/style-analyzer/pull/654

I continue rewriting things using annotations.


zurk:
Also closing because it is based on closed PR.

PR #709, opened 2019-03-15
Add GitHub suggestion comments generation

EgorBu:
* template
* logic to generate several candidates per identifier
* logic to compute probabilities

Now it's possible to generate comments like:
<details>

## Example output 1
Possible typos in `Token_Taken_Tuken`, suggestions:
Candidate 1 with confidence 33%
```suggestion
var Token_Tokens_Tokens;
```
Candidate 2 with confidence 33%
```suggestion
var Token_Tokens_Then;
```
Candidate 3 with confidence 32%
```suggestion
var Token_Tokens_Token;
```

## Example output 2
Possible typos in `someWrang_Variuble_NOME`, suggestions:
Candidate 1 with confidence 33%
```suggestion
var someRange_Variable_NODE = "bla";
```
Candidate 2 with confidence 33%
```suggestion
var someWrap_Variable_NODE = "bla";
```
Candidate 3 with confidence 33%
```suggestion
var someLang_Variable_NODE = "bla";
```

## Example output 3
Possible typos in `some_wrang_variuble_nome`, suggestions:
Candidate 1 with confidence 33%
```suggestion
var some_range_variable_node = "bla";
```
Candidate 2 with confidence 33%
```suggestion
var some_wrap_variable_node = "bla";
```
Candidate 3 with confidence 33%
```suggestion
var some_lang_variable_node = "bla";
```
</details>

### DIFF
--- a/lookout/style/typos/analyzer.py
+++ b/lookout/style/typos/analyzer.py
@@ -160,14 +160,15 @@ class IdTyposAnalyzer(Analyzer):
                     identifier_candidates = [
                         Candidate(i, c) for i, c in zip(
                             sugg_identifiers, self._normalize_confidences(id_confidences),
-                        )
+                        ) if i != identifier
                     ]
-                    yield TypoFix(
-                        head_file=file,
-                        identifier=identifier,
-                        candidates=candidates,
-                        line_number=new_identifiers[index].start_position.line,
-                        identifier_candidates=identifier_candidates)
+                    if identifier_candidates:
+                        yield TypoFix(
+                            head_file=file,
+                            identifier=identifier,
+                            candidates=candidates,
+                            line_number=new_identifiers[index].start_position.line,
+                            identifier_candidates=identifier_candidates)
 
     def render_comment_text(self, typo_fix: TypoFix) -> str:
         """
--- a/lookout/style/typos/analyzer.py
+++ b/lookout/style/typos/analyzer.py
@@ -1,6 +1,8 @@
 """Identifier typos analyzer."""
 from collections import defaultdict
+import itertools
 import logging
+import os
 from typing import Any, Dict, Iterable, Iterator, List, Mapping, NamedTuple, Sequence, Tuple
 
 import bblfsh
@@ -10,19 +12,21 @@ from lookout.core.data_requests import DataService, \
     with_changed_uasts_and_contents, with_uasts_and_contents
 from lookout.core.lib import extract_changed_nodes, files_by_language, filter_files, find_new_lines
 from lookout.sdk.service_data_pb2 import Change, File
+import numpy
 import pandas
 from sourced.ml.algorithms import TokenParser, uast2sequence
 
-from lookout.style.common import merge_dicts
+from lookout.style.common import load_jinja2_template, merge_dicts
 from lookout.style.format.utils import generate_comment
 from lookout.style.typos.corrector_manager import TyposCorrectorManager
-from lookout.style.typos.utils import Candidate, Columns, flatten_df_by_column
+from lookout.style.typos.utils import Candidate, Columns, flatten_df_by_column, TEMPLATE_DIR
 
 TypoFix = NamedTuple("LineFix", (
     ("head_file", File),  # file from head revision
     ("line_number", int),  # line number for the comment
-    ("token", str),  # token where typo is found
-    ("candidates", List[Candidate]),  # Candidates for token fix
+    ("identifier", str),  # identifier where typo is found
+    ("candidates", Mapping[str, Iterable[Candidate]]),  # Candidates for token {token: candidates}
+    ("identifier_candidates", Iterable[Candidate]),  # Suggested identifiers
 ))
 
 
@@ -38,6 +42,7 @@ class IdTyposAnalyzer(Analyzer):
     version = 1
     description = "Corrector of typos in source code identifiers."
     corrector_manager = TyposCorrectorManager()
+    vendor = "source{d}"
 
     default_config = {
         "line_length_limit": 500,
@@ -46,6 +51,7 @@ class IdTyposAnalyzer(Analyzer):
         "overall_size_limit": 5 << 20,  # 5 MB
         "model": "245fae3a-2f87-4990-ab9a-c463393cfe51",
         "index_column": "index",
+        "comment_template": os.path.join(TEMPLATE_DIR, "comment.md.jinja2"),
     }
 
     def __init__(self, model: AnalyzerModel, url: str, config: Mapping[str, Any]):
@@ -60,6 +66,7 @@ class IdTyposAnalyzer(Analyzer):
         self.config = self._load_config(config)
         self.model = self.corrector_manager.get(self.config["model"])
         self.parser = self.create_token_parser()
+        self.comment_template = load_jinja2_template(self.config["comment_template"])
 
     @staticmethod
     def create_token_parser() -> TokenParser:
@@ -87,12 +94,18 @@ class IdTyposAnalyzer(Analyzer):
         :return: List of found review suggestions. Refer to \
                  lookout/core/server/sdk/service_analyzer.proto.
         """
-        return [generate_comment(
-            filename=typo_fix.head_file.path,
-            line=typo_fix.line_number,
-            text=self.render_comment_text(typo_fix),
-            confidence=self._get_comment_confidence(typo_fix.candidates))
-            for typo_fix in self.generate_typos_fixes(list(changes))]
+        comments = []
+        for typo_fix in self.generate_typos_fixes(list(changes)):
+            text = self.render_comment_text(typo_fix)
+            confidence = 1 - numpy.prod(
+                1 - numpy.fromiter((x.confidence for x in typo_fix.identifier_candidates),
+                                   dtype=numpy.float),
+            )
+
+            comments.append(generate_comment(text=text, confidence=int(100 * confidence),
+                                             filename=typo_fix.head_file.path,
+                                             line=typo_fix.line_number))
+        return comments
 
     def generate_typos_fixes(self, changes: Sequence[Change]) -> Iterator[TypoFix]:
         """
@@ -132,15 +145,29 @@ class IdTyposAnalyzer(Analyzer):
                 self._log.debug("Found %d new identifiers, first one: %s" %
                                 (len(new_identifiers), new_identifiers[0].token))
                 suggestions = self.check_identifiers([n.token for n in new_identifiers])
+                if not suggestions:
+                    continue
                 for index in suggestions.keys():
-                    corrections = suggestions[index]
-                    for token in corrections.keys():
-                        yield TypoFix(
-                            head_file=file,
-                            token=new_identifiers[index].token,
-                            candidates=corrections[token],
-                            line_number=new_identifiers[index].start_position.line,
+                    identifier = new_identifiers[index].token
+                    candidates = {token: [Candidate(*sugg) for sugg in suggestions[index][token]]
+                                  for token in suggestions[index]}
+                    sugg_identifiers, id_confidences = [], []
+                    for final_sugg, conf in self.generate_identifier_suggestions(
+                            candidates, identifier):
+                        sugg_identifiers.append(final_sugg)
+                        id_confidences.append(conf)
+
+                    identifier_candidates = [
+                        Candidate(i, c) for i, c in zip(
+                            sugg_identifiers, self._normalize_confidences(id_confidences),
                         )
+                    ]
+                    yield TypoFix(
+                        head_file=file,
+                        identifier=identifier,
+                        candidates=candidates,
+                        line_number=new_identifiers[index].start_position.line,
+                        identifier_candidates=identifier_candidates)
 
     def render_comment_text(self, typo_fix: TypoFix) -> str:
         """
@@ -149,17 +176,56 @@ class IdTyposAnalyzer(Analyzer):
         :param typo_fix: Information about typo fix required to render a comment text.
         :return: string with the generated comment.
         """
-        # TODO: move this logic to template. See FormatAnalyzer.render_comment_text()
-        corrections_line = ", ".join("%s (%d%%)" % (
-            c.token, int(c.confidence * 100)) for c in typo_fix.candidates)
-        return "Possible typo in \"%s\". Suggestions: %s" % (typo_fix.token, corrections_line)
+        file_lines = typo_fix.head_file.content.decode("utf-8", "replace").splitlines()
+        confidences = [candidate.confidence for candidate in typo_fix.identifier_candidates]
+        return self.comment_template.render(
+            identifier=typo_fix.identifier,
+            suggestions=[candidate.token for candidate in typo_fix.identifier_candidates],
+            old_code_line=file_lines[typo_fix.line_number - 1],
+            confidences=confidences,
+            zip=zip)
 
     @staticmethod
-    def _get_comment_confidence(suggestions: Sequence[Candidate]) -> int:
-        invert_confidence = 1
-        for suggestion in suggestions:
-            invert_confidence *= (1 - suggestion.confidence)
-        return int(100 * (1 - invert_confidence))
+    def _normalize_confidences(confidences: Sequence[float]):
+        s = sum(confidences)
+        return [conf / s for conf in confidences]
+
+    def generate_identifier_suggestions(self, suggestions: Mapping[str, Iterable[Candidate]],
+                                        identifier: str) \
+            -> Iterator[Tuple[str, float]]:
+        """
+        Generate suggestions for the identifier and compute the probability of suggestion.
+
+        :param suggestions: suggestions are a mapping from a token to the list of candidates.
+        :param identifier: initial identifier.
+        :return: a generator of tuples with a suggestion for the identifier and probability.
+        """
+        if not suggestions:
+            # do nothing in case of empty suggestions
+            return
+        token_candidates = []
+        # split identifier
+        initial_tokens = self.parser.split(identifier)
+
+        # prepare suggestions per initial token
+        for t in initial_tokens:
+            token_candidates.append(suggestions[t])
+
+        # sort candidates by resulting identifier probability
+        suggestion_candidates = list(reversed(sorted(itertools.product(*token_candidates),
+                                                     key=self._proba)))
+
+        for i in range(min(self.config["n_candidates"], len(suggestion_candidates))):
+            final_candidate = self.reconstruct_identifier(
+                tokenizer=self.parser,
+                pred_tokens=[c.token for c in suggestion_candidates[i]],
+                identifier=identifier,
+            )
+            yield final_candidate, self._proba(suggestion_candidates[i])
+
+    @staticmethod
+    def _proba(candidates: Iterable[Candidate]):
+        return numpy.prod([c.confidence for c in candidates])
 
     @staticmethod
     def reconstruct_identifier(tokenizer: TokenParser, pred_tokens: List[str], identifier: str) \
@@ -173,7 +239,7 @@ class IdTyposAnalyzer(Analyzer):
         :return: reconstructed identifier based on predicted tokens.
         """
         identifier_l = identifier.lower()
-        # setup required parameters
+        # check required parameters
         assert tokenizer._single_shot, "TokenParser should be initialized with " \
                                        "`single_shot=True` for IdTyposAnalyzer"
         # sanity checking
@@ -238,7 +304,7 @@ class IdTyposAnalyzer(Analyzer):
         df[Columns.Split] = [" ".join(self.parser.split(i)) for i in identifiers]
         df = flatten_df_by_column(df, Columns.Split, Columns.Token, str.split)
         suggestions = self.model.suggest(df, n_candidates=self.config["n_candidates"],
-                                         return_all=False)
+                                         return_all=True)
         suggestions = self.filter_suggestions(df, suggestions)
         grouped_suggestions = defaultdict(dict)
         for index, row in df.iterrows():
--- a/lookout/style/typos/templates/comment.md.jinja2
+++ b/lookout/style/typos/templates/comment.md.jinja2
@@ -1,0 +1,7 @@
+Possible typo in `{{ identifier }}`, fixes:
+{% for suggestion, confidence in zip(suggestions, confidences) %}
+Candidate #{{ loop.index }} with a confidence {{ (100 * confidence) | int }}%
+```suggestion
+{{ old_code_line.replace(identifier, suggestion) }}
+```
+{% endfor %}
--- a/lookout/style/typos/utils.py
+++ b/lookout/style/typos/utils.py
@@ -1,6 +1,7 @@
 """Various glue functions to work with the input dataset and the output from FastText."""
 import csv
 from itertools import chain
+import os
 from typing import Dict, Iterable, List, NamedTuple, Optional, Set, Tuple
 
 import numpy
@@ -21,6 +22,8 @@ Candidate = NamedTuple("Candidate", (
     ("token", str),         # Token with fixed typo
     ("confidence", float),  # Model confidence for the correction
 ))
+
+TEMPLATE_DIR = os.path.join(os.path.dirname(__file__), "templates")
 
 
 def filter_splits(data: pandas.DataFrame, tokens: Set[str]) -> pandas.DataFrame:
--- a/lookout/style/typos/metrics.py
+++ b/lookout/style/typos/metrics.py
@@ -6,7 +6,7 @@ import pandas
 from sklearn.metrics import accuracy_score, precision_recall_fscore_support
 
 from lookout.style.common import load_jinja2_template
-from lookout.style.typos.utils import Columns
+from lookout.style.typos.utils import Columns, TEMPLATE_DIR
 
 
 @unique
@@ -71,6 +71,5 @@ def get_scores(data: pandas.DataFrame, suggestions: Dict[int, List[Tuple[str, fl
 def generate_report(data: pandas.DataFrame, suggestions: Dict[int, List[Tuple[str, float]]],
                     ) -> str:
     """Print scores for suggestions in an easy readable way."""
-    template = load_jinja2_template(
-        os.path.join(os.path.dirname(__file__), "templates", "scores.md.jinja2"))
+    template = load_jinja2_template(os.path.join(TEMPLATE_DIR, "scores.md.jinja2"))
     return template.render(ScoreMode=ScoreMode, get_scores=get_scores, **locals())
--- a/lookout/style/typos/tests/test_analyzer.py
+++ b/lookout/style/typos/tests/test_analyzer.py
@@ -64,6 +64,11 @@ class AnalyzerTests(unittest.TestCase):
                 model=MODEL_PATH, confidence_threshold=0.0, n_candidates=3))
         comments = analyzer.analyze(ptr_from=self.ptr, ptr_to=self.ptr, data_service=dataservice)
         self.assertGreater(len(comments), 0)
+        bad_names = ["nam", "print_tipe", "gett_lenght"]
+        good_names = ["name", "print_type", "get_length"]
+        for c in comments:
+            self.assertFalse(any(name in c.text.split(", fixes:")[0] for name in good_names))
+            self.assertTrue(any(name in c.text.split(", fixes:")[0] for name in bad_names))
 
     def test_reconstruct_identifier(self):
         tokens = [


lookout/style/typos/analyzer.py
```diff
@@ -145,17 +155,69 @@ def render_comment_text(self, typo_fix: TypoFix) -> str:
...
+        final_sugg = []
+        confidences = [1]
+        suggestion_queue = []
+        for i, t in enumerate(initial_tokens):
+            try:
+                final_sugg.append(suggestions[t][0].token)
+                confidences.append(suggestions[t][0].confidence)
+                for sugg in suggestions[t][1:]:
+                    suggestion_queue.append((i, sugg.token, sugg.confidence))
+            except KeyError:
```
zurk: I'd like to have something like:
```python
if t in suggestions:
    ...
else:
    ...
```
instead of try-except. It makes logic move straightforward.

lookout/style/typos/analyzer.py
```diff
@@ -145,17 +155,69 @@ def render_comment_text(self, typo_fix: TypoFix) -> str:
...
+        yield final_sugg, self._compute_confidence(confidences)
+
+        suggestion_queue = list(reversed(sorted(suggestion_queue, key=itemgetter(2))))
+        for j in range(n_candidates - 1):
+            pos = suggestion_queue[j][0]
+            sugg = suggestion_queue[j][1]
+            conf = suggestion_queue[j][2]
+            final_sugg[pos] = sugg
+            confidences[pos] = conf
+            yield final_sugg, self._compute_confidence(confidences)
```
zurk: I spent quite a lot of time trying to understand the logic of the last (203-220) lines.
I believe there should be a more straightforward way to do the same. I think that `suggestions` have tokens for the provided identifier only, so we can use `product` from itertools. https://docs.python.org/3/library/itertools.html#itertools.product to do the same. 

So I'd suggest you iterate through the loop of the product of changes suggestions.  

And a question. Does suggestion list contain an original token with a probability (so a suggestion to not change this token)?

lookout/style/typos/templates/comment.md.jinja2
```diff
@@ -0,0 +1,8 @@
+format: typos:
+Possible typos in `{{ identifier }}`, suggestions:
+{% for i, (suggestion, confidence) in enumerate(zip(suggestions, confidences)) %}
+Candidate {{ i + 1 }} with confidence {{ int(100 * confidence) }}%
```
zurk:
```suggestion
Candidate #{{ i + 1 }} with a confidence {{ int(100 * confidence) }}%
```

lookout/style/typos/analyzer.py
```diff
@@ -90,8 +97,9 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
             filename=typo_fix.head_file.path,
             line=typo_fix.line_number,
             text=self.render_comment_text(typo_fix),
-            confidence=self._get_comment_confidence(typo_fix.candidates))
-            for typo_fix in self.generate_typos_fixes(list(changes))]
+            confidence=1)
```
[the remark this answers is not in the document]
EgorBu: **First**, the formula is not correct - if we have one subtoken with probability 1 - the total probability will be 1 even with small probabilities for other subtokens.
**Second**, for independent events, probability is computed `p_ev1 * p_ev2 * p_ev3 * ... p_evK` [wiki](https://en.wikipedia.org/wiki/Independence_(probability_theory)#More_than_two_events)
**Third**, one comment here contains several suggestions with computed probability like I put in the PR description:
## Example output 1
Possible typos in `Token_Taken_Tuken`, suggestions:
Candidate 1 with confidence 33%
```suggestion
var Token_Tokens_Tokens;
```
Candidate 2 with confidence 33%
```suggestion
var Token_Tokens_Then;
```
Candidate 3 with confidence 32%
```suggestion
var Token_Tokens_Token;
```

lookout/style/typos/analyzer.py
```diff
@@ -145,17 +155,69 @@ def render_comment_text(self, typo_fix: TypoFix) -> str:
...
-    def _get_comment_confidence(suggestions: Sequence[Candidate]) -> int:
-        invert_confidence = 1
-        for suggestion in suggestions:
-            invert_confidence *= (1 - suggestion.confidence)
-        return int(100 * (1 - invert_confidence))
+    def _normalize_confidences(confidences: Sequence[float]):
+        s = sum(confidences)
+        return [conf / s for conf in confidences]
+
+    def generate_identifier_suggestions(self, suggestions: Mapping[str, Iterable[Candidate]],
```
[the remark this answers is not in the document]
EgorBu: I prefer to keep this argument - it makes code less spaghetti style.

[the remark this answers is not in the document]
zurk: `@staticmethod` then?

lookout/style/typos/analyzer.py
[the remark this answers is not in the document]
zurk: > First, the formula is not correct - if we have one subtoken with probability 1 - the total probability will be 1 even with small probabilities for other subtokens.

It is correct. What do I calculate by this formula? I calculate `p` the probability of getting the correct suggestion in the list of provided suggestions (that is what I understand by comment confidence). We can put `p` `p = 1 - p_{neg}` where `p_{neg}` is the probability of all comments be wrong. If `p_{i}` is the probability of `i`-th comment to be correct that means that `1-p_{i}` is the probability of `i`-th comment to be wrong. Applying the fact that we assume suggestions independent (please refer the link to the wiki you added) we get
`p_{neg} = (1- p_1) * ... (1 - p_n)`

And so 
`p = 1 - (1- p_1) * ... (1 - p_n)`

> Second, for independent events, probability is computed

I know thank you.

> Third, one comment here contains several suggestions with computed probability like I put in the PR description:

Yes, I saw. It is a different number and different probabilities. 

----
Please, Egor, if you disagree, or think your opponent is wrong, do not be so pushy next time.
If you see any negative from my side please do tell me.

lookout/style/typos/analyzer.py
EgorBu: A couple of points:
We have quite a lot of places with the same logic to check `key`
[style analyzer](https://github.com/src-d/style-analyzer/search?q=keyerror&unscoped_q=keyerror)
[src-d/ml](https://github.com/src-d/ml/search?q=keyerror&unscoped_q=keyerror)

And several more arguments https://stackoverflow.com/questions/12342822/python-try-except-keyerror-vs-if-has-key?lq=1
TLDR:
it's cheaper to have `exception` in case of rare `KeyErrors` than check `if condition` on every iteration.
To make a proper decision we need to evaluate it on real data to understand if it's worth to switch logic to checking `if condition` at each iteration (frequent `KeyErrors`).

lookout/style/typos/analyzer.py
EgorBu: `itertools.product` will have high computational complexity - `n_tokens ^ n_candidates` + sorting them to select best candidates so using big-O notation will give use:
* `O(product + sorting)` -> `O(n_tokens ^ n_candidates + n_tokens ^ n_candidates * log(n_tokens ^ n_candidates))`.

The current solution has 
* `O(select_best_candidates + sort_rest_candidates)` -> `O(n_tokens + n_tokens * n_candidates * log(n_tokens * n_candidates))` 

So the current solution is much better in terms of computational complexity.

lookout/style/typos/analyzer.py
```diff
@@ -145,17 +155,69 @@ def render_comment_text(self, typo_fix: TypoFix) -> str:
...
+                                        identifier: str, n_candidates: int = 3) \
+            -> Iterator[Tuple[List[str], float]]:
+        """
+        Generate suggestions for identifier and compute probability of suggestion.
+
+        :param suggestions: suggestions - mapping from token to list of candidates.
+        :param identifier: initial identifier.
```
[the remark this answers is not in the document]
EgorBu: We don't know if it has a typo or not.

lookout/style/typos/analyzer.py
zurk: I do not like that it hides possible bugs if `suggestions[t]` is empty. 
I do not see performance issues here. 
if you want to keep try-except let's add assert

EgorBu: It's handled here https://github.com/src-d/style-analyzer/blob/f00b3eeac4ff0085b4a73c76cc7827b2172d9e87/lookout/style/typos/analyzer.py#L139~L140
How `if key in suggestions` can solve an empty `suggestions` issue?

lookout/style/typos/analyzer.py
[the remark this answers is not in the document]
EgorBu: After discussion - it was a misunderstanding of `token`/`identifier` probabilities.
TLDR: we have a suggestion that contains several possible identifiers. Ex:
```javascript
Token_Tokens_Tokens
Token_Tokens_Then
Token_Tokens_Token
```
We have probability for each identifier in suggestion, we need to compute confidence for suggestion.
@zurk suggested to use probabilities of identifiers (not tokens as I misunderstood initially)

lookout/style/typos/analyzer.py
zurk: it did not hide the error and you get an exception.
let's just add an assert. I think it is fine.

lookout/style/typos/analyzer.py
zurk: ok, can you add some comments, because now I am not sure I understand the logic?

EgorBu: About logic - read it carefully - if you still have a question - call me.
About question - @irinakhismatullina knows the answer.

zurk: Should everyone call you who do not understand the logic? ;)
It is good to have comments for every unobvious code. Because you increase code reading speed and it is imprortant.

irinakhismatullina: The original token is not present yet. Before it would be like *why the hell are you suggesting me to change this tokens to itself*. So i think we should add it, but as a separate info, not like one of the candidates. I'll think of the best way to do it. How urgent is it?

vmarkovtsev: \+1 for adding a comment about the logic in any case

zurk: Thanks, @irinakhismatullina!
What we need is an ability to get a probability that token is correct and should not be fixed. Otherwise, we can have such a situation: 

Identifier: TupoThen
Suggestions: 
1. Tupo -> Typo (with prob 0.9)
2. Then -> Than (with prop 0.1)

So we do not have an option to change only `Tupo` and keep `Then`.

The initial token itself we can get from https://github.com/src-d/style-analyzer/pull/709/files#diff-18972d7c14aaaa30c48c1560b2e149c0R204

irinakhismatullina: In the example you provided most likely the corrector would not suggest to correct `Then` at all (it makes only the suggestions that have higher proba, than the proba of the token itself). Anyway this must be added, but used carefully.

vmarkovtsev: \-1 for being noisy in our predictions

EgorBu: https://github.com/src-d/style-analyzer/blob/c93cb3147d604fa1be7da7a763963cb5e78b56f3/lookout/style/typos/analyzer.py#L190~L223

lookout/style/typos/analyzer.py
```diff
@@ -145,17 +151,114 @@ def render_comment_text(self, typo_fix: TypoFix) -> str:
...
+        `final_sugg` :  [`same`, `range`, `for`]
+        `confidences`: [0.9, 0.8, 0.8]
+        4) put other candidates in a queue where the first element in tuple is position in
+           `final_sugg`
+        [(0, `some`, 0.8), (1, `wrong`, 0.7), (2, `var`, 0.6), (1, `lang`, 0.5)]
+        and sort queue by probability
+        5) generate next identifier suggestions using sorted candidates -
+        ------------------------------------
+            next candidate: (0, `some`, 0.8)
+            `final_sugg` :  [`some`, `range`, `for`]
```
irinakhismatullina: I don't understand why you keep modifying `final_sugg` with every new candidate? So in the end all additional candidates have less probable `"some"` instead of `"same"`, which seems definitely wrong to me. Imo `final_sugg` should be fixed in the form you returned first, and then replace one token just for one yield,  while leaving `final_sugg`  as it is as a standard.

EgorBu: good point. In this case @zurk suggestion about `itertools.product` should be used.

lookout/style/typos/analyzer.py
zurk: > it makes only the suggestions that have higher proba, than the proba of the token itself

ok, good!

lookout/style/typos/utils.py
```diff
@@ -22,6 +23,8 @@
     ("confidence", float),  # Model confidence for the correction
 ))
 
+TEMPLATE_DIR = os.path.join(os.path.dirname(__file__), "templates")
```
zurk: I believe it is not in `__init__.py` as we have in `lookout.style.format` because of some import problems, right?

lookout/style/typos/analyzer.py
```diff
@@ -128,34 +136,85 @@ def generate_typos_fixes(self, changes: Sequence[Change]) -> Iterator[TypoFix]:
...
-                            head_file=file,
-                            token=new_identifiers[index].token,
-                            candidates=corrections[token],
-                            line_number=new_identifiers[index].start_position.line,
-                        )
-
-    def render_comment_text(self, typo_fix: TypoFix) -> str:
+                    yield TypoFix(
+                        head_file=file,
+                        identifier=new_identifiers[index].token,
```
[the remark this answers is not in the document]
irinakhismatullina: It's an identifier, babelfish notation: `new_itentifiers[index]` - Node, `Node.token` - identifier itself.

lookout/style/typos/utils.py
EgorBu: Circular import.
In format analyzer, it's solved in configuration https://github.com/src-d/style-analyzer/blob/master/lookout/style/format/config.py#L75
So I'm not sure if it's used anywhere https://github.com/src-d/style-analyzer/blob/master/lookout/style/format/__init__.py#L8

lookout/style/typos/analyzer.py
```diff
@@ -128,34 +136,85 @@ def generate_typos_fixes(self, changes: Sequence[Change]) -> Iterator[TypoFix]:
...
         """
-        # TODO: move this logic to template. See FormatAnalyzer.render_comment_text()
-        corrections_line = ", ".join("%s (%d%%)" % (
-            c.token, int(c.confidence * 100)) for c in typo_fix.candidates)
-        return "Possible typo in \"%s\". Suggestions: %s" % (typo_fix.token, corrections_line)
+        file_lines = typo_fix.head_file.content.decode("utf-8", "replace").splitlines()
+        suggestions, confidences = [], []
+        for final_sugg, conf in self.generate_identifier_suggestions(
+                typo_fix.candidates, typo_fix.identifier,
+        ):
```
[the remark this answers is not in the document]
EgorBu: I don't think so. 
You mentioned `_get_comment_confidence` in comments above - and it was called in `render_comment_text` - so it's a logic specific for comment generation. As everything inside `render_comment_text` is a logic specific for the generation of specific types of comments (1 loop only now).
`typos_fixes` already contains all the required information to generate everything. But obviously, it contains only generic information.
Before it was subtoken to subtoken suggestions. Now - identifier to identifier.
And it shouldn't be bloated with non-generic information IMHO because it will increase the complexity of very `lookout-specific` function `generate_typos_fixes` and we will lose in flexibility and extendability.

lookout/style/typos/analyzer.py
```diff
@@ -86,12 +91,15 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
-        return [generate_comment(
-            filename=typo_fix.head_file.path,
-            line=typo_fix.line_number,
-            text=self.render_comment_text(typo_fix),
-            confidence=self._get_comment_confidence(typo_fix.candidates))
-            for typo_fix in self.generate_typos_fixes(list(changes))]
+        comments = []
+        for typo_fix in self.generate_typos_fixes(list(changes)):
+            text, confidences = self.render_comment_text(typo_fix)
+            confidence = 1 - reduce(lambda x, y: x * y, map(lambda x: 1 - x, confidences))
```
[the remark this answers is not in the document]
EgorBu: commented below

lookout/style/typos/utils.py
zurk: yeah, looks like we should do something similar in style-analyzer.

lookout/style/typos/analyzer.py
zurk: I think that was missed ^

lookout/style/typos/analyzer.py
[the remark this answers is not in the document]
zurk: Sorry for being unclear, @EgorBu, the point is that I need to call this logic from the reporter to get identifiers suggestion if I do not know what do we exactly recommend. Let's agree that reporter should not do it even if it is two lines logic.

So, the `analyze` function split to `generate_typos_fixes` and `render_comment_text` was by design created to have everything that reporter need in `TypoFix`. That why I asked to put in inside `generate_typos_fixes()`.

lookout/style/typos/analyzer.py
zurk: ok, outdated.

lookout/style/typos/templates/comment.md.jinja2
```diff
@@ -0,0 +1,8 @@
+format: typos:
```
irinakhismatullina: Why is is exactly `format: typos:`? Didn't you mean `style: typos:`? Won't there be any confusion between `format` and `typos` analyzers?

lookout/style/typos/analyzer.py
```diff
@@ -128,34 +136,85 @@ def generate_typos_fixes(self, changes: Sequence[Change]) -> Iterator[TypoFix]:
...
+            return
+        token_candidates = []
+        # split identifier
+        initial_tokens = self.parser.split(identifier)
+
+        # prepare suggestions per initial token
+        for t in initial_tokens:
+            try:
+                token_candidates.append(suggestions[t])
+            except KeyError:
```
irinakhismatullina: Btw, you can avoid it setting `return_all=True` in TyposCorrector calling [here](https://github.com/src-d/style-analyzer/pull/709/files#diff-18972d7c14aaaa30c48c1560b2e149c0R295). This way suggestions will contain all tokens, and the ones that are considered not-typoed will have candidates in this form: `A: [A, 1.0]` - exactly what you do here. In your way there are much less information passed, so maybe it's better to leave it as it is, idk.

lookout/style/typos/templates/comment.md.jinja2
EgorBu: yep, WDYT about `typos: misspelled identifiers correction:`?

EgorBu:
```
commit 9683f33fec085533738dc839731405aae846e7f6 (origin/add-token, add-token)
Author: Irina <irenekhismatullina@gmail.com>
Date:   Wed Mar 20 18:08:27 2019 +0300

    Adjust analyzer to show the token proba, if present

    Signed-off-by: Irina <irenekhismatullina@gmail.com>

commit c1b45efb0ea1dbabcbdf713974dfa91b6da5017c
Author: Irina <irenekhismatullina@gmail.com>
Date:   Wed Mar 20 17:58:56 2019 +0300

    Always add token to candidates

    Signed-off-by: Irina <irenekhismatullina@gmail.com>
```
Couple of commits breaks the logic - waiting for updates from @irinakhismatullina to rebase.

lookout/style/typos/analyzer.py
```diff
@@ -13,16 +16,17 @@
...
-from lookout.style.typos.utils import Candidate, Columns, flatten_df_by_column
+from lookout.style.typos.utils import Candidate, Columns, flatten_df_by_column, TEMPLATE_DIR
 
 TypoFix = NamedTuple("LineFix", (
     ("head_file", File),  # file from head revision
     ("line_number", int),  # line number for the comment
-    ("token", str),  # token where typo is found
-    ("candidates", List[Candidate]),  # Candidates for token fix
+    ("identifier", str),  # identifier where typo is found
+    ("candidates", Mapping[str, Iterable[Candidate]]),  # Candidates for token fix
```
zurk: let's make it `Sequence` because if you want to iterate more then 1 time it is better.

Also not sure about our typing agreement
Should we just specify exact types like `Dict[str, List[Candidate]]`?

@vmarkovtsev WDYT?

lookout/style/typos/analyzer.py
```diff
@@ -13,16 +16,17 @@
...
+    ("identifiers", Iterable[Candidate]),  # Suggested identifiers
```
zurk: Also `Sequence`

lookout/style/typos/analyzer.py
```diff
@@ -129,34 +139,91 @@ def generate_typos_fixes(self, changes: Sequence[Change]) -> Iterator[TypoFix]:
...
-                            token=new_identifiers[index].token,
-                            candidates=corrections[token],
-                            line_number=new_identifiers[index].start_position.line,
+                    identifier = new_identifiers[index].token
+                    candidates = {token: [Candidate(*sugg) for sugg in suggestions[index][token]]
+                                  for token in suggestions[index]}
+                    sugg_identifiers, ident_confidences = [], []
+                    for final_sugg, conf in self.generate_identifier_suggestions(
+                            candidates, identifier,
+                    ):
```
zurk: let's merge to the previous line?
Now `):` fits the limit.

lookout/style/typos/analyzer.py
zurk: Also, It is a little bit confusing to have `identifier` and `identifiers`, don't you think?
Maybe, let's rename `candidates ` to `token_candidates` and `identifiers` to `identifier_candidates`?

lookout/style/typos/analyzer.py
```diff
@@ -86,12 +93,15 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
         """
-        return [generate_comment(
-            filename=typo_fix.head_file.path,
-            line=typo_fix.line_number,
-            text=self.render_comment_text(typo_fix),
-            confidence=self._get_comment_confidence(typo_fix.candidates))
-            for typo_fix in self.generate_typos_fixes(list(changes))]
+        comments = []
+        for typo_fix in self.generate_typos_fixes(list(changes)):
+            text, confidences = self.render_comment_text(typo_fix)
```
zurk: I believe that now you do not need to return confidences from `render_comment_text` because you already have them in `typo_fix`. Let's return text only. 

In this case, the old code works with a small change:
```python
return [generate_comment(
            filename=typo_fix.head_file.path,
            line=typo_fix.line_number,
            text=self.render_comment_text(typo_fix),
            confidence=self._get_comment_confidence(typo_fix. identifiers))
            for typo_fix in self.generate_typos_fixes(list(changes))]
```
Up to you to take it back  or use your map-reduce formula :)

lookout/style/typos/analyzer.py
EgorBu: @zurk, you mixing definitions of `Iterable` and `Generator` - [lmgtfy](https://nvie.com/posts/iterators-vs-generators/) in case of 1 time iterations. 
If you can find access by index or checking the length of `Iterable` - it can be narrowed down to `Sequence`. If no - no. And the same answer for similar suggestions.

lookout/style/typos/analyzer.py
EgorBu: `identifier_candidates` :+1: 
`Sequence` :-1:

lookout/style/typos/analyzer.py
zurk: Yes, thanks for the pic. 
I still disagree because it is a List by fact because we set it to the list below, but ok, whatever.

vmarkovtsev: Can you please add a comment about what is the mapping key and what are the iterable values

lookout/style/typos/analyzer.py
```diff
@@ -86,12 +93,16 @@ def analyze(self, ptr_from: ReferencePointer, ptr_to: ReferencePointer,
...
-        return [generate_comment(
-            filename=typo_fix.head_file.path,
-            line=typo_fix.line_number,
-            text=self.render_comment_text(typo_fix),
-            confidence=self._get_comment_confidence(typo_fix.candidates))
-            for typo_fix in self.generate_typos_fixes(list(changes))]
+        comments = []
+        for typo_fix in self.generate_typos_fixes(list(changes)):
+            text = self.render_comment_text(typo_fix)
+            confidence = 1 - reduce(lambda x, y: x * y, map(lambda x: 1 - x.confidence,
```
vmarkovtsev: Much clearer goal:
```suggestion
            confidence = 1 - numpy.prod(1 - numpy.fromiter(x.confidence for x in  typo_fix.identifier_candidates))
```

lookout/style/typos/analyzer.py
```diff
@@ -129,15 +140,29 @@ def generate_typos_fixes(self, changes: Sequence[Change]) -> Iterator[TypoFix]:
...
-                    for token in corrections.keys():
-                        yield TypoFix(
-                            head_file=file,
-                            token=new_identifiers[index].token,
-                            candidates=corrections[token],
-                            line_number=new_identifiers[index].start_position.line,
+                    identifier = new_identifiers[index].token
+                    candidates = {token: [Candidate(*sugg) for sugg in suggestions[index][token]]
+                                  for token in suggestions[index]}
+                    sugg_identifiers, ident_confidences = [], []
```
vmarkovtsev: sugg?
\+
```suggestion
                    sugg_identifiers, id_confidences = [], []
```

lookout/style/typos/analyzer.py
```diff
@@ -146,17 +171,57 @@ def render_comment_text(self, typo_fix: TypoFix) -> str:
...
-        return "Possible typo in \"%s\". Suggestions: %s" % (typo_fix.token, corrections_line)
+        file_lines = typo_fix.head_file.content.decode("utf-8", "replace").splitlines()
+        confidences = [candidate.confidence for candidate in typo_fix.identifier_candidates]
+        return self.comment_template.render(
+            identifier=typo_fix.identifier,
+            suggestions=[candidate.token for candidate in typo_fix.identifier_candidates],
+            old_code_line=file_lines[typo_fix.line_number - 1],
+            confidences=confidences,
+            zip=zip,
+            enumerate=enumerate,
```
vmarkovtsev: `enumerate` is a built-in. http://jinja.pocoo.org/docs/2.10/templates/#for -> `loop.index`

lookout/style/typos/analyzer.py
```diff
@@ -146,17 +171,57 @@ def render_comment_text(self, typo_fix: TypoFix) -> str:
...
+            int=int)
```
vmarkovtsev: `int` is already there. `"4556" | int`

lookout/style/typos/analyzer.py
```diff
@@ -1,6 +1,9 @@
 """Identifier typos analyzer."""
 from collections import defaultdict
+from functools import reduce
+from itertools import product
```
vmarkovtsev: Can you please use `itertools.product`, because "product" alone may mean many things. E.g. what we are building right now :)

lookout/style/typos/analyzer.py
```diff
@@ -146,17 +171,57 @@ def render_comment_text(self, typo_fix: TypoFix) -> str:
...
+        token_candidates = []
+        # split identifier
+        initial_tokens = self.parser.split(identifier)
+
+        # prepare suggestions per initial token
+        for t in initial_tokens:
+            token_candidates.append(suggestions[t])
+
+        # sort candidates by resulting identifier probability
+        suggestion_candidates = list(reversed(sorted(product(*token_candidates), key=self._proba)))
```
vmarkovtsev: 👍 smart

lookout/style/typos/analyzer.py
```diff
@@ -146,17 +171,57 @@ def render_comment_text(self, typo_fix: TypoFix) -> str:
...
+            yield final_candidate, self._proba(suggestion_candidates[i])
 
     @staticmethod
-    def _get_comment_confidence(suggestions: Sequence[Candidate]) -> int:
-        invert_confidence = 1
-        for suggestion in suggestions:
-            invert_confidence *= (1 - suggestion.confidence)
-        return int(100 * (1 - invert_confidence))
+    def _proba(candidates: Iterable[Candidate]):
+        return reduce(lambda x, y: x * y, [c.confidence for c in candidates])
```
vmarkovtsev:
```suggestion
        return numpy.prod([c.confidence for c in candidates])
```

lookout/style/typos/templates/comment.md.jinja2
```diff
@@ -0,0 +1,8 @@
+typos: misspelled identifiers correction:
```
vmarkovtsev:
```suggestion
Typo in `{{ identifier }}`, fixes:
```

lookout/style/typos/templates/comment.md.jinja2
```diff
@@ -0,0 +1,8 @@
...
+Possible typos in `{{ identifier }}`, suggestions:
+{% for i, (suggestion, confidence) in enumerate(zip(suggestions, confidences)) %}
+Candidate #{{ i + 1 }} with a confidence {{ int(100 * confidence) }}%
```
vmarkovtsev:
```suggestion
Candidate #{{ loop.index }} with a confidence {{ 100 * confidence | int }}%
```

Note: `loop.index0` is 0-based, and `loop.index` is 1-based. Jinja2 is awesome.

lookout/style/typos/analyzer.py
EgorBu: `sugg` - suggested

lookout/style/typos/templates/comment.md.jinja2
EgorBu: I would mention `Possible` because it's not 100% accurate

EgorBu:
Test added

lookout/style/typos/templates/comment.md.jinja2
vmarkovtsev: Ok

zurk:
@vmarkovtsev let's merge?

PR #710, opened 2019-03-18
Deploy both format and typos to staging

meyskens:
This PR makes sure both analyzers run individually inside staging


meyskens:
@vmarkovtsev thanks! will await working further on this till that is added

vmarkovtsev:
OK, @irinakhismatullina has added the missing feature, so it should work without including the asdf file now.

meyskens:
@vmarkovtsev @irinakhismatullina do we need any configuration for the analyzer in lookout now (`--config-json='{"lookout.style.typos":{"/src-d/style-analyzer/corrector.asdf"}}'`) still? Or is there a default model now?

vmarkovtsev:
@meyskens There is a default model now indeed, so no.

zurk:
@meyskens I think it is worth to mention that now we use JS driver v2.7.1 (not v.1.2.0 as before) for style Analyzer and Typos Analyzer required all possible drivers.
@vmarkovtsev do you think we should pin versions for the rest drivers now?

meyskens:
@zurk addressed in https://github.com/src-d/lookout/pull/619

meyskens:
This request is now ready for review! Once ready to deploy a new tagged version is needed to be pushed.

vmarkovtsev:
OK, we are very close to the initial typos release, so pushing the tag very soon. Thanks!

PR #717, opened 2019-03-21
Fix `TyposCorrector` creation

EgorBu:

### DIFF
--- a/lookout/style/typos/corrector_manager.py
+++ b/lookout/style/typos/corrector_manager.py
@@ -4,6 +4,8 @@ to be used globally.
 """
 from typing import Iterable
 
+import modelforge.backends
+
 from lookout.style.typos.corrector import TyposCorrector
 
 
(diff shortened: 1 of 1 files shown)


lookout/style/typos/corrector_manager.py
```diff
@@ -25,7 +27,7 @@ def get(self, source: str) -> TyposCorrector:
         """
         obj = self.correctors.get(source)
         if obj is None:
-            obj = TyposCorrector().load(source)
+            obj = TyposCorrector().load(source, backend=modelforge.backends.create_backend())
```
zurk: We need it to make it work if `source` is some part of  `UUID`, right?

vmarkovtsev: Yes, it should work with UUIDs. Actually, I need to edit Modelforge to create the backend inside, so this will not be needed.

PR #723, opened 2019-03-26
Improve ignore lists

m09:


lookout/style/format/visualizer/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules
+models/model.asdf
```
zurk: why not all directory?

m09: only model.asdf is used so normally there is no reason to put any other file in it.

PR #729, opened 2019-03-27
Polish IdTyposAnalyzer

zurk:
Minor changes to simplify work with an upcoming reporter.
1. The biggest limitation I am addressing is that `TypoFix` instance should be easily converted to a JSON object. So I exclude UAST from it.
2. Use `@with_changed_uasts_and_contents(unicode=True)`
3. Convert `pred_probs[i]` to float because it has `np.float` type which fails be converted to a JSON.
4. Remove `candidates` to `token_candidates` to be more specific.

### DIFF
--- a/lookout/style/typos/analyzer.py
+++ b/lookout/style/typos/analyzer.py
@@ -6,12 +6,11 @@ import os
 from typing import Any, Dict, Iterable, Iterator, List, Mapping, NamedTuple, Sequence, Tuple
 
 import bblfsh
-from lookout.core.analyzer import Analyzer, AnalyzerModel, DummyAnalyzerModel, ReferencePointer
+from lookout.core.analyzer import Analyzer, AnalyzerModel, DummyAnalyzerModel, ReferencePointer, \
+    UnicodeChange
 from lookout.core.api.service_analyzer_pb2 import Comment
-from lookout.core.data_requests import DataService, \
-    with_changed_uasts_and_contents, with_uasts_and_contents
+from lookout.core.data_requests import DataService, with_changed_uasts_and_contents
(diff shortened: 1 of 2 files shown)


lookout/style/typos/analyzer.py
```diff
@@ -21,12 +20,14 @@
...
-    ("candidates", Mapping[str, Iterable[Candidate]]),  # Candidates for token {token: candidates}
-    ("identifier_candidates", Iterable[Candidate]),  # Suggested identifiers
+TypoFix = NamedTuple("TypoFix", (
+    ("content", str),                                         # file content from head revision
+    ("path", str),                                            # file path from head revision
+    ("line_number", int),                                     # line number for the comment
+    ("identifier", str),                                      # identifier where typo is found
+    ("token_candidates", Mapping[str, Iterable[Candidate]]),  # candidates for token
+                                                              # {token: candidates}
+    ("identifier_candidates", Iterable[Candidate]),           # suggested identifiers
```
vmarkovtsev: Can you build `identifier_candidates` from `token_candidates`? If you can, then better to inherit and implement the getter.

zurk: well, I actually do not need `"token_candidates", let's just drop it and keep `identifier_candidates` only?

vmarkovtsev: That's what the original code was before somebody added the mapping. OK, let's remove then

vmarkovtsev: And name it "canidates"

zurk:
addressed and rebased.

vmarkovtsev:
@zurk Please merge yourself if CI passes

PR #730, opened 2019-03-28
Remove logging args from args given to handlers

m09:


m09:
ok. Then there must be some problem in setup.py because my version was not updated even though I ran `pip install .`

m09:
they don't seem removed: https://github.com/src-d/lookout-sdk-ml/blob/master/lookout/__main__.py#L7

vmarkovtsev:
I am talking about https://github.com/src-d/modelforge/blob/master/modelforge/slogging.py#L251

vmarkovtsev:
@m09 I bet that Modelforge is not updated on your side because of the version conflict with sourced-ml. This is why we had to hack the CI, and also opened https://github.com/src-d/ml/issues/396

m09:
yup ok, let's close then

PR #732, opened 2019-03-28
Fix parent bug

### LINKED ISSUE #663: FeatureExtractor._fill_vnode_parents() bug
Sometimes you get a parent UAST node for the vnode wrongly.
Example:
```
AssertionError: 
Wrong parent of j=3 
node=VirtualNode(' Polyfill ES6 Promise', y=None, start=(2, 1, 3), end=(23, 1, 24), node=4747332400, path="test/specs/__helpers.js"). 
parent.start_position.offset=24, parent.end_position.offset=58
```
So parent does not cover original UAST Node by offsets.

m09:
Fix #663. As mentioned during the sync, there were two bugs in our code but also the general invariant "A parent covers its children" is not respected in the js driver (comments can be outside of their parent).

### DIFF
--- a/lookout/style/format/feature_extractor.py
+++ b/lookout/style/format/feature_extractor.py
@@ -803,7 +803,8 @@ class FeatureExtractor:
         for j, vn in enumerate(file_vnodes):
             if vn.node:
                 closest_left_node_id = id(vn.node)
-            parent = self._find_parent(j, file_vnodes, file_parents, closest_left_node_id)
-            if parent is None:
-                parent = uast
-            vnode_parents[id(vn)] = parent
+                vnode_parents[id(vn)] = file_parents[id(vn.node)]
+            else:
+                parent = (self._find_parent(j, file_vnodes, file_parents, closest_left_node_id)
+                          or uast)
+                vnode_parents[id(vn)] = parent
--- a/lookout/style/format/uast_stability_checker.py
+++ b/lookout/style/format/uast_stability_checker.py
@@ -101,7 +101,7 @@ class UASTStabilityChecker:
             if vnode.start.offset not in start_offset_to_vnodes:
                 # NOOP always included
                 start_offset_to_vnodes[vnode.start.offset] = i
-        for i, vnode in enumerate(vnodes[::-1]):
+        for i, vnode in enumerate(vnodes[::-1], start=1):
             if vnode.end.offset not in end_offset_to_vnodes:
                 # NOOP always included that is why we have reverse order in this loop
                 end_offset_to_vnodes[vnode.end.offset] = len(vnodes) - i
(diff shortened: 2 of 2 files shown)


lookout/style/format/feature_extractor.py
```diff
@@ -803,7 +803,8 @@ def _fill_vnode_parents(self, file_parents: Mapping[int, bblfsh.Node],
...
                 closest_left_node_id = id(vn.node)
-            parent = self._find_parent(j, file_vnodes, file_parents, closest_left_node_id)
-            if parent is None:
-                parent = uast
-            vnode_parents[id(vn)] = parent
+                vnode_parents[id(vn)] = file_parents[id(vn.node)]
+            else:
+                parent = (self._find_parent(j, file_vnodes, file_parents, closest_left_node_id)
+                          or uast)
+                vnode_parents[id(vn)] = parent
```
zurk: From the code, it is hard to get how exactly you fix the problem. Can you elaborate a bit?

lookout/style/format/uast_stability_checker.py
```diff
@@ -141,7 +141,6 @@ def _check_file(
             vnode_start_index = start_offset_to_vnodes[start]
             vnode_end_index = end_offset_to_vnodes[end]
 
-            assert vnode_start_index <= vnodes_i < vnode_end_index
```
zurk: Let's keep it just in case. It is harmless.

zurk: Or is it triggered because
> also the general invariant "A parent covers its children" is not respected in the js driver

? 

Then we should fix `UastStabilityChecker` because it uses this assumption. 
In such a case, it is another problem for another PR then, I think

m09: It is triggered indeed.

lookout/style/format/feature_extractor.py
m09: I fix 2 other parent bugs but those do not fix the driver problem itself --- it remains.

lookout/style/format/uast_stability_checker.py
zurk: let's keep an assert because it means that we still should address it properly.
I think I know how.

m09: I'd rather fix UastStabilityChecker not to rely on this invariant

vmarkovtsev: whatever you decide, I am fine with both options

zurk: Ok, let's keep it as it is now, I created an issue to do not lose it. https://github.com/src-d/style-analyzer/issues/733

PR #735, opened 2019-03-29
Store old identifiers in the IdTyposModel

irinakhismatullina:
1. Create IdTyposModel to store known identifiers after training.
2. Small optimizations in `check_identifiers()`:
- Don't check the same identifier several times
- Use `return_all=False`. Before comments were generated for all new identifiers, now only for corrected ones. We could still do it with `return_all=True`, but it would be more difficult, and more data would be passed, we don't need that.
3. Add `min_token_length` parameter, for now it's turned off, but can be used in the future (to not check one-letter tokens, for example).


lookout/style/typos/analyzer.py
```diff
@@ -207,8 +254,9 @@ def generate_identifier_suggestions(self, suggestions: Mapping[str, Iterable[Can
         initial_tokens = self.parser.split(identifier)
 
         # prepare suggestions per initial token
-        for t in initial_tokens:
-            token_candidates.append(suggestions[t])
+        for token in initial_tokens:
+            #
```
vmarkovtsev: ?

lookout/style/typos/analyzer.py
```diff
@@ -273,20 +324,36 @@ def reconstruct_identifier(tokenizer: TokenParser, pred_tokens: List[str], ident
...
+        _log.info("train %s %s %s %s", __version__, ptr.url, ptr.commit,
+                  pformat(train_config, width=4096, compact=True))
+        model = IdTyposModel()
+        for _, files in files_by_language(files).items():
+            for file in filter_files(
+                    files=files, line_length_limit=train_config["line_length_limit"],
+                    overall_size_limit=train_config["overall_size_limit"], log=_log):
+                model.update({
+                    node.token for node in uast2sequence(file.uast)
+                    if bblfsh.role_id("IDENTIFIER") in node.roles
```
vmarkovtsev: `bblfsh.role_id("IDENTIFIER")` should be at top function level - it takes time for every call. The same about `IMPORT`

lookout/style/typos/analyzer.py
```diff
@@ -297,18 +364,25 @@ def check_identifiers(self, identifiers: List[str],
         :return: Dictionary of corrections grouped by ids of corresponding identifier \
                  in 'identifiers' and typoed tokens which have correction suggestions.
         """
-        df = pandas.DataFrame(columns=[self.config["index_column"], Columns.Split])
-        df[self.config["index_column"]] = range(len(identifiers))
-        df[Columns.Split] = [" ".join(self.parser.split(i)) for i in identifiers]
+        identifiers_positions = defaultdict(list)
+        for i, identifier in identifiers:
+            identifiers_positions[identifier].append(i)
+        unique_identifiers = list(identifiers_positions.keys())
```
vmarkovtsev:
```suggestion
        unique_identifiers = sorted(identifiers_positions.keys())
```

vmarkovtsev:
@irinakhismatullina There is a conflict after merging @zurk's PR

vmarkovtsev:
@irinakhismatullina No conflicts now, but the CI failed

irinakhismatullina:
Yep, working on it. Something with the reporter tests, we're figuring it out with @zurk

lookout/style/typos/analyzer.py
```diff
@@ -284,20 +336,37 @@ def reconstruct_identifier(tokenizer: TokenParser, pred_tokens: List[str], ident
         return "".join(res)
 
     @classmethod
+    @with_uasts_and_contents(unicode=True)
```
zurk: Should be False. Not all languages can be correctly transformed now. 
```suggestion
    @with_uasts_and_contents(unicode=False)
```

lookout/style/typos/analyzer.py
```diff
@@ -284,20 +336,37 @@ def reconstruct_identifier(tokenizer: TokenParser, pred_tokens: List[str], ident
...
+                  pformat(train_config, width=4096, compact=True))
+        model = IdTyposModel()
+        if not train_config["store_identifiers"]:
+            return model
+        for _, files in files_by_language(files).items():
+            for file in filter_files(
+                    files=files, line_length_limit=train_config["line_length_limit"],
+                    overall_size_limit=train_config["overall_size_limit"], log=_log):
+                model.update({
+                    node.token for node in uast2sequence(file.uast)
```
zurk: I think that it is possible to use `self._get_identifiers()` here because they both do the same thing.
```suggestion
                    node.token for node in self._get_identifiers(file.uast, [])
```

Is it correct?

lookout/style/typos/analyzer.py
```diff
@@ -308,18 +377,25 @@ def check_identifiers(self, identifiers: List[str],
...
         """
-        df = pandas.DataFrame(columns=[self.config["index_column"], Columns.Split])
-        df[self.config["index_column"]] = range(len(identifiers))
-        df[Columns.Split] = [" ".join(self.parser.split(i)) for i in identifiers]
+        identifiers_positions = defaultdict(list)
+        for i, identifier in enumerate(identifiers):
+            identifiers_positions[identifier].append(i)
+        unique_identifiers = sorted(identifiers_positions.keys())
+        df = pandas.DataFrame(columns=[IDENTIFIER_ID_COLUMN, Columns.Split])
+        df[IDENTIFIER_ID_COLUMN] = range(len(unique_identifiers))
```
zurk: well, you have `df.index` with the same values, I believe. 
Do you need this column or you can use `df.index` instead?

lookout/style/typos/analyzer.py
irinakhismatullina: Yes, it should be

lookout/style/typos/analyzer.py
irinakhismatullina: No, I call `flatten` afterwards (to have one token per row instead of one identifier per row) and I need to store this `identifier id` separately to be able to combine suggestions for the tokens inside the same identifier.

lookout/style/typos/analyzer.py
```diff
@@ -284,20 +292,35 @@ def reconstruct_identifier(tokenizer: TokenParser, pred_tokens: List[str], ident
...
+        _log.info("train %s %s %s %s", __version__, ptr.url, ptr.commit,
+                  pformat(train_config, width=4096, compact=True))
+        model = IdTyposModel()
+        for _, files in files_by_language(files).items():
+            for file in filter_files(
+                    files=files, line_length_limit=train_config["line_length_limit"],
+                    overall_size_limit=train_config["overall_size_limit"], log=_log):
+                model.identifiers.update({
+                    node.token for node in cls._get_identifiers(file.uast, [])
+                    if IDENTIFIER in node.roles and IMPORT not in node.roles and node.token
```
zurk:
```suggestion
```
because checks are in `_get_identifiers` already

irinakhismatullina: Yep, thanks

irinakhismatullina:
@vmarkovtsev Added tests to check that the feature works, ready to review/merge.

lookout/style/typos/analyzer.py
```diff
@@ -40,50 +38,7 @@
...
-
-    def __len__(self) -> int:
-        return len(self._identifiers)
-
-    def __iter__(self):
-        yield from self._identifiers.__iter__()
-
-    def __contains__(self, identifier: str) -> bool:
-        return identifier in self._identifiers
+IDENTIFIER_ID_COLUMN = "identifier_id"
```
vmarkovtsev: If you unbox the abbreviation, it means "identifier_identifer". Let's rename to "identifier_index", if it **is** an index

lookout/style/typos/analyzer.py
```diff
@@ -122,9 +78,10 @@ def __init__(self, model: IdTyposModel, url: str, config: Mapping[str, Any]):
         self.corrector = self.corrector_manager.get(self.config["corrector"])
         self.parser = self.create_token_parser()
         self.comment_template = load_jinja2_template(self.config["comment_template"])
-        self.known_identifiers = model
-        for identifier in self.known_identifiers:
+        for identifier in model.identifiers:
             self.corrector.expand_vocabulary(set(self.parser.split(identifier)))
+        self.approved_identifiers = set() if self.config["check_all_identifiers"] else \
```
vmarkovtsev: approved -> allowed

PR #740, opened 2019-04-03
Fix UASTFixer reliance on non-existing invariant of parent covers children

### LINKED ISSUE #733: Fix UastStabilityChecker
Right now `UastStabilityChecker` relies on the fact that "A parent covers its children" in the UAST (See discussion https://github.com/src-d/style-analyzer/pull/732#discussion_r270110695) 
It is not true and we should address it properly.

m09:
Fix #733.

### DIFF
--- a/lookout/style/format/uast_stability_checker.py
+++ b/lookout/style/format/uast_stability_checker.py
@@ -39,7 +39,7 @@ class UASTStabilityChecker:
         self._parsing_cache = {}  # type: Dict[int, Optional[Tuple[bblfsh.Node, int, int]]]
         self._debug = debug
 
-    def _parse_code(self, parent: bblfsh.Node, content: str,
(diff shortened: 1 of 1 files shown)


lookout/style/format/uast_stability_checker.py
```diff
@@ -69,11 +70,12 @@ def _parse_code(self, parent: bblfsh.Node, content: str,
             descendants.append(current_ancestor)
             start, end = (current_ancestor.start_position.offset,
                           current_ancestor.end_position.offset)
-            uast, errors = parse_uast(stub, content[start:end], filename="", unicode=True,
-                                      language=self._feature_extractor.language)
-            if not errors:
-                result = uast, start, end
-                break
+            if end and start <= vnode.start.offset and end > vnode.end.offset:
```
zurk: why not just 
```python
if start <= vnode.start.offset and end > vnode.end.offset:
```
(without `end` check)?

PR #745, opened 2019-04-09
Add evaluation on identifiers

irinakhismatullina:
1. Add evaluation on solo identifiers.
2. Add Candidates class usage in some more missed places.
3. Add `"processes number:` parameter to the analyzer config.

PR #751, opened 2019-04-10
Fix formatting for typos report

zurk:
Tabulate cannot work with different formatting for the same column, that why I manually format all values.
I create a function for this because of some jijla2 limitations. You cannot run arbitrary python code there that output nothing.


lookout/style/typos/benchmarks/typo_commits_report.py
```diff
@@ -247,6 +247,14 @@ def _finalize(self, reports: Iterable[Dict[str, str]]) -> Iterator[Dict[str, str
         :param reports: Reports generated by `TypoCommitsReporter.generate_commit_dataset_report()`
         :return: Summarized final report
         """
+        def format_series(series: pandas.Series, formatting: Tuple[str, str]) -> pandas.Series:
+            series = series.copy()
+            keys = []
+            for key, fmt in formatting:
+                keys.append(key)
+                series.loc[key] = ("%" + fmt) % scores[key]
```
vmarkovtsev: @Guillemdb Is there a clever `map` for this kind of operation in pandas?

Guillemdb: yes, using applymap on a dataframe/map on a series passing a lambda function, but depending on what series and scores contains it can be made even faster.

Just to make sure I understand the code:

What is supposed to be contained in series? (index and value).
What is exactly `scores`? it looks like a series. what is the index and value?the indexes are in the same order as `series`?

The returned value of the function what is it supposed to be?

If you pass me a toy example of inputs -> outputs I can benchmark the fastest approach in pandas.

zurk: `scores` is a series, yes and unfortunately there is no simple way to convert series with map-style function if conversion depends on the index. 

@Guillemdb this code should give you an understanding what is inside `scores`.
https://github.com/src-d/style-analyzer/blob/5ecd2d5709160087bdb9a0c368a7b4e956953624/lookout/style/typos/benchmarks/typo_commits_report.py#L279-L290

What I want to do is to change the order and convert values depending on the index value. If you can suggest something better then loop I glad to use it.

Guillemdb: You are right, that is a use case that requires a bit of dark magic.

```python
def format_series_pd(series: pd.Series, formatting: Tuple[str, str]) -> pd.Series:
    # Create a DataFrame for the formatting so we can work with it in pandas
    f_ix, f_vals = zip(*formatting)
    formatting = pd.DataFrame(data=list(f_vals), index=list(f_ix), columns=["format"])
    df = pd.merge(left=formatting,
                  right=scores.reset_index(), # Merge works only on DataFrames
                  left_index=True,
                  right_on="index")
    df = df.set_index("index")
    func_to_apply = lambda x: "{}{}".format(x["scores"], x["format"])
    return df.apply(func_to_apply, axis=1) # Apply row-wise```

As you can see, the output is the same. Please confirm me that the format tuples have a correct format.

![dataframe_konst](https://user-images.githubusercontent.com/7149899/55941459-904be280-5c42-11e9-8a67-8717ec4f2e3d.png)

zurk: well, I prefer this tiny loop :)

Guillemdb: Yeah, in your case using the loop it is worth it, specially if the number of metrics its not very big.  Pandas is adding too much overhead for such a simple calculation.

zurk:
Looks like I hit the same thing
https://github.com/src-d/style-analyzer/pull/745#issuecomment-481719842 in this PR and others
so I am waiting for merge of #745.

lookout/style/typos/benchmarks/typo_commits_report.py
vmarkovtsev: OK

zurk:
@vmarkovtsev PTAL. Tests are good now.

PR #756, opened 2019-04-12
Add reports for format analyzer release

zurk:
Noisy report in pending


zurk:
Noisy report is a little bit strange but I suggest to address it in v0.2.1 if we found that something is wrong.

zurk:
CI passed, it is just a glitch

PR #759, opened 2019-04-22
Explicitly add xz support to smart_open

irinakhismatullina:
The guys decided to roll back and remove `.xz` support, added by @vmarkovtsev.


vmarkovtsev:
We are going to make a hard fork and avoid this. Please pin the exact package version for now.

PR #760, opened 2019-04-22
Pin the xz-supporting version of smart-open

irinakhismatullina:


setup.py
```diff
@@ -48,7 +48,7 @@
         "xgboost>=0.72,<2.0",
         "tabulate>=0.8.0,<2.0",
         "python-igraph>=0.7.0,<2.0",
-        "smart-open>=1.8.1,<2.0",
+        "smart-open>=1.8.1,<1.8.2",
```
zurk:
```suggestion
        "smart-open==1.8.1",
```

PR #763, opened 2019-05-15
Add filtered dataset and code to do the filtering

irinakhismatullina:
1. Filter items, where token splits of the wrong and the correct identifiers are equal (they differ in non-letter symbols or case spelling).
2. Filter items, where wrong and correct identifiers are equal on lemmas level.

The result is obtained with the updated [TokenParser](https://github.com/src-d/ml/blob/master/sourced/ml/algorithms/token_parser.py), which is __not included in the last src-d/ml release__. Therefore, to have results reproducible by default, we either need to make a release, or load the package from the github on setup.


irinakhismatullina:
Btw src-d/ml-core has [old TokenParser](https://github.com/src-d/ml-core/blob/master/sourced/ml/core/algorithms/token_parser.py), so right now it cannot be used.

zurk:
Ok, I will update it.

zurk:
PR with an update: https://github.com/src-d/ml-core/pull/13

lookout/style/typos/research/eval_dataset/filter_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+
+import pandas
+import spacy
+
+from sourced.ml.algorithms import TokenParser
+
+
+def filter_test_dataset(dataset: str, filtered_dataset: str) -> None:
+    """
+    Filter non-typo-ed identifiers from the dataset.
```
vmarkovtsev:
```suggestion
    Extract non-typo-ed identifiers from the dataset.
```
Someone has to write a linter rule to disable repeating the function name parts in the docstring.

lookout/style/typos/research/eval_dataset/filter_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+
+    1. Filter items, where token splits of the wrong and the correct identifiers are equal (they differ in non-letter \
```
vmarkovtsev:
```suggestion
    1. Filter items, where token splits of the wrong and the correct identifiers are equal (they differ in non-alpha \
```

lookout/style/typos/research/eval_dataset/filter_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+    symbols or case spelling).
```
vmarkovtsev:
```suggestion
    chars or casing).
```

lookout/style/typos/research/eval_dataset/filter_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+    2. Filter items, where wrong and correct identifiers are equal on lemmas level.
+    :param dataset: Path to the test dataset.
```
vmarkovtsev: **test** dataset?

lookout/style/typos/research/eval_dataset/filter_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+
+    # Filter examples with equal splits
+    tp = TokenParser(min_split_length=1, stem_threshold=400, single_shot=True, max_token_length=400,
+                     attach_upper=True)
+    data["wrong_split"] = data["wrong"].apply(lambda x: " ".join(tp.split(x)))
+    data["correct_split"] = data["correct"].apply(lambda x: " ".join(tp.split(x)))
+    data = data[data["wrong_split"] != data["correct_split"]]
+
+    os.system("python3 -m spacy download en")
+    nlp = spacy.load('en', disable=['parser', 'ner'])
```
vmarkovtsev: Single quotes.

lookout/style/typos/research/eval_dataset/filter_dataset.py
irinakhismatullina: Don't you think that it's cool if the function name really reflects the best way you would describe it's work in a text?
Anyway I thought of the better name which gives clearer understanding of what the function does.

vmarkovtsev: Totally cool, that's right. Not cool to have a docstring which does not give any additional information about the function, that's my point :)

vmarkovtsev: I particularly don't like the word "filter" in source code because it can mean different things:

1. Remove something and return the rest.
2. Extract something, return it, and discard the rest.

Those are two opposite actions. Different programmers all over the world keep using "filter" at random and it really puzzles every time when you read their code. This is why repeating "filter" in the docstring is so unhelpful.

irinakhismatullina: Totally agree, I think I myself sometimes use it for both, and always feel confused, so i really like the new name:)

lookout/style/typos/research/eval_dataset/filter_dataset.py
```diff
@@ -0,0 +1,45 @@
...
+import spacy
+
+from sourced.ml.algorithms import TokenParser
+
+
+def remove_non_typos(dataset: str, filtered_dataset: str) -> None:
+    """
+    Remove non-typo-ed identifiers from the dataset.
+
+    1. Remove examples, where token splits of the wrong and the correct identifiers are equal (they differ in non-alpha \
```
vmarkovtsev: This line is 120 characters. How come that it passed the CI?

vmarkovtsev: Ah, because we do not check `research` for formatting. We should change this.

PR #766, opened 2019-05-20
Fix cloner tests 

zurk:
This commit address the recent changes in modelforge (https://github.com/src-d/modelforge/commit/9e10ff214ad356d4073ebadd6374df961ee88975)

The test expected to get each log entry at one line, but it is not the case now.

@vmarkovtsev actually, I am not sure that it is a good idea to have JSON logs formatted with new lines. JSON logs are for processing by design and not for humans. In other words, I think it is better to revert commit in modelforge rather than apply this fix.

### DIFF
--- a/lookout/style/tests/test_cloner.py
+++ b/lookout/style/tests/test_cloner.py
@@ -45,8 +45,10 @@ class ClonerTests(unittest.TestCase):
             "failed to clone",
             "successfully cloned 5/6 repositories",
         ]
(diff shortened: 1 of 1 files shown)


vmarkovtsev:
Oh f*ck this is my fault, I missed that @glimow has changed slogging additionally to the required edits. Yes, we need to remove the slogging indentation ofc.

zurk:
https://github.com/src-d/modelforge/pull/98

Issue #768, opened 2019-05-21
Run the basic formatting checks over the research code


vmarkovtsev:
We should run the basic PEP8 checks over the code in `research`. `pycodestyle --max-line-length=99`, no more than that.

PR #769, opened 2019-05-22
Fix style in research directories

zurk:
Closes https://github.com/src-d/style-analyzer/issues/768

Also I add a simple check `pycodestyle --max-line-length=99` for research directories


vmarkovtsev:
Awesome. We should not make the same mistake twice and check PEP8 in the future research projects.

PR #772, opened 2019-05-24
Change fasttext training config

irinakhismatullina:
1. Change in the model parameters.
2. No corruption of training data. Corruption sometimes improves the quality, but not dramatically. Without it the model size is much lower.

### DIFF
--- a/lookout/style/typos/config.py
+++ b/lookout/style/typos/config.py
@@ -18,12 +18,12 @@ DEFAULT_CORRECTOR_CONFIG = {
         "prepared_filename": "prepared.csv",
     },
     "fasttext": {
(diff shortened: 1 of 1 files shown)


zurk:
can you give some insights about why do you change this number?  
I believe the new parameters improve quality in one of the reports?

irinakhismatullina:
The most important change is `corrupt=False`. As I said above, it's change significantly reduces the model's size, and there's almost no quality drop.
The other ones are for the quality improvement in this setup in comparison with the default one with only `corrupt` changed to `False`.

I can't promise that it's the best configuration, but it's the one I was happy with after a series of experiments, and the one I used in all my training experiments, and in the final model. So mostly they are exactly the same to reproduce my results.

PR #784, opened 2019-06-21
Small change in fasttext training data generation

irinakhismatullina:
- Use real identifier frequencies for sampling - more fair distribution
- Use 2-tokens identifiers too, before it started only with 3

### DIFF
--- a/lookout/style/typos/config.py
+++ b/lookout/style/typos/config.py
@@ -28,6 +28,7 @@ DEFAULT_CORRECTOR_CONFIG = {
         "path": str(DEFAULT_DATA_DIR / "fasttext.bin"),  # Where to store trained fasttext model
         "dim": 10,  # Number of dimensions of embeddings
         "bucket": 200000,  # Number of hash buckets in the model
+        "adjust_frequencies": True,
     },
     "datasets": {
         "portion": 400000,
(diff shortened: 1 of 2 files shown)


lookout/style/typos/preparation.py
```diff
@@ -260,8 +260,9 @@ def train_fasttext(data: pandas.DataFrame, config: Optional[Mapping[str, Any]] =
     if config is None:
         config = {}
     config = merge_dicts(DEFAULT_CORRECTOR_CONFIG["fasttext"], config)
-    train_data = data[[len(str(x).split()) > 2 for x in data[Columns.Split]]].sample(
-        config["size"], weights=Columns.Frequency, replace=True)
+    lengths = data[Columns.Split].apply(lambda x: len(str(x).split()))
```
zurk:
```suggestion
    tokens_number = data[Columns.Split].apply(lambda x: len(str(x).split()))
```
I think it fits more.

irinakhismatullina:
It doesn't affect performance significantly afaik, but this way the distribution for sample is more fair, like more correct. Actually this division by the number of tokens makes sense only for some data format, that is used by default, and probably it should be made a function argument.
WDYT?

zurk:
let's put it to `config`.

lookout/style/typos/config.py
```diff
@@ -28,6 +28,7 @@
         "path": str(DEFAULT_DATA_DIR / "fasttext.bin"),  # Where to store trained fasttext model
         "dim": 10,  # Number of dimensions of embeddings
         "bucket": 200000,  # Number of hash buckets in the model
+        "adjust_frequencies": True,
```
zurk:
```suggestion
        "adjust_frequencies": True,  # Whether to divide identifiers frequencies by tokens number.
```
Please adjust the comment if I get it wrongly.

Issue #754, opened 2019-04-11
Update gensim.fasttext usage


irinakhismatullina:
New release of `gensim==3.7.2` broke several things in `CandidateGenerator` in `typos`:

1. Loading fasttext model from the file - _easy fix_.
2. Modelforge save/load functions are now broken because of the changed attributes in `FasttextKeyedVectors` class - _not nice, need to rewrite the_ `_generate_tree/_load_tree`, _but doable_.
3. Because of (2) new models for usage and testing should be trained, and that's _the worst of all_ (reports inconsistencies and so on).

So due to (3) it's suggested to leave it for the future releases, when the new models will be integrated anyway.

zurk:
@irinakhismatullina I think this one was done, wasn't it?

irinakhismatullina:
Nope, we just put upper boundary on the gensim version.

PR #785, opened 2019-06-21
Refactor generation and update gensim.fasttext usage

irinakhismatullina:
- Support the last version of gensim #754 
- Change generation hyperparameters
- Small optimizations (less edit distance calculations, less lookups in sets)


lookout/style/typos/generation.py
```diff
@@ -109,7 +108,7 @@ def set_config(self, config: Optional[Mapping[str, Any]] = None) -> None:
         """
         if config is None:
             config = {}
-        self.config = merge_dicts(DEFAULT_CORRECTOR_CONFIG["generation"], config)
+        self.config = merge_dicts(self.config, config)
```
zurk:
```suggestion
        self._config = merge_dicts(self._config, config)
```

lookout/style/typos/generation.py
```diff
@@ -198,41 +197,44 @@ def _lookup_corrections_for_token(self, typo_info: TypoInfo) -> List[Features]:
         candidates = []
         candidate_tokens = self._get_candidate_tokens(typo_info)
         typo_vec = self._vec(typo_info.typo)
-        dist_calc = EditDistance(typo_info.typo, "damerau")
-        for candidate in set(candidate_tokens):
-            candidate_vec = self.wv[candidate]
-            dist = dist_calc.damerau_levenshtein_distance(candidate, self.config["radius"])
+        for candidate, dist in set(candidate_tokens):
```
zurk: `candidate_tokens` is set already 
```suggestion
        for candidate, dist in candidate_tokens:
```

lookout/style/typos/generation.py
```diff
@@ -198,41 +197,44 @@ def _lookup_corrections_for_token(self, typo_info: TypoInfo) -> List[Features]:
...
                 if edit_candidates_count >= self.config["edit_dist_number"]:
                     continue
-                candidate_tokens.append(suggestion.term)
+                candidate_tokens.add((suggestion.term, suggestion.distance))
                 edit_candidates_count += 1
         if self.config["neighbors_number"] > 0:
             typo_neighbors = self._closest(self._vec(typo_info.typo),
                                            self.config["neighbors_number"])
-            candidate_tokens.extend(typo_neighbors)
+            candidate_tokens |= set([(
```
zurk: You can use a generator directly here. So I suggest removing additional `[]` or replace them with `()` if it does not work without brackets.

lookout/style/typos/generation.py
```diff
@@ -198,41 +197,44 @@ def _lookup_corrections_for_token(self, typo_info: TypoInfo) -> List[Features]:
...
+                candidate,
+                dist_calc.damerau_levenshtein_distance(candidate, self.config["radius"]))
```
zurk: I like more how it was before: `_get_candidate_tokens()` returns a list of tokens and no additional information. 

Now you should think carefully and look in code what the second number means. Let's revert this change.

lookout/style/typos/generation.py
```diff
@@ -293,7 +295,7 @@ def _vec(self, token: str) -> numpy.ndarray:
         return self.wv[token]
 
     def _freq(self, token: str) -> float:
-        return float(self.frequencies.get(token, self.min_freq))
+        return float(self.frequencies.get(token, 0))
```
zurk: was it a bug?

lookout/style/typos/generation.py
```diff
@@ -375,18 +377,25 @@ class DummyModel(Model):
...
+            buckets_word_lengths = numpy.zeros(len(self.wv.buckets_word), dtype=numpy.uint32)
+            buckets_word_values = []
+            for word_index in range(len(self.wv.buckets_word)):
+                buckets = self.wv.buckets_word[word_index]
+                buckets_word_lengths[word_index] = len(buckets)
+                buckets_word_values.extend(sorted(buckets))
+            buckets_word = {"lengths": buckets_word_lengths,
+                            "values": numpy.array(buckets_word_values)}
+        else:
+            buckets_word = numpy.array([])
```
zurk: `buckets_word` has a different type: `dict` (in if branch) and `numpy.array` in else. Looks like a bug.

Should be something like 
```python
buckets_word = {"lengths": numpy.array([]),
                "values": numpy.array([])}
```

lookout/style/typos/generation.py
```diff
@@ -419,9 +428,15 @@ def _load_tree(self, tree: dict) -> None:
             for i, s in enumerate(vocab)}
         wv.bucket = self.wv["bucket"]
         wv.index2word = wv.index2entity = vocab
-        wv.num_ngram_vectors = self.wv["num_ngram_vectors"]
         wv.vectors_ngrams = numpy.array(self.wv["vectors_ngrams"])
-        wv.hash2index = {k: v for v, k in enumerate(self.wv["hash2index"])}
+        if "buckets_word" in self.wv.keys() and isinstance(self.wv["buckets_word"], dict):
```
zurk: `"buckets_word" in self.wv.keys()`. `buckets_word` should be always in a tree.

lookout/style/typos/generation.py
```diff
@@ -1,9 +1,9 @@
 """Generation of the typo correction candidates. Contains features extraction and serialization."""
 from itertools import chain
 from multiprocessing import Pool
-from typing import Any, Iterable, List, Mapping, NamedTuple, Optional, Set, Union
+from typing import Any, Iterable, List, Mapping, NamedTuple, Optional, Set, Tuple, Union
 
-from gensim.models import FastText
+from gensim.models.fasttext import load_facebook_vectors
```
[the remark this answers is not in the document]
irinakhismatullina: What do you mean? It's a method for loading vectors that were saved with the original fasttext implementation (by facebook), and that's exactly what I need. Idk why they chose exactly that name¯\_(ツ)_/¯

lookout/style/typos/generation.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, **kwargs):
         self.wv = None
         self.tokens = set()
         self.frequencies = {}
-        self.min_freq = 0
+        self.config = DEFAULT_CORRECTOR_CONFIG["generation"]
```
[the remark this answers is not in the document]
irinakhismatullina: Generator is a part of corrector, and its config is a part of the whole corrector config.

lookout/style/typos/generation.py
irinakhismatullina: The thing is, the distance calculation is a costly operation, so why repeat it twice

lookout/style/typos/generation.py
irinakhismatullina: No, it was a feature. In new conditions using 0 works better for quality.

lookout/style/typos/generation.py
irinakhismatullina: No, it is not for the previous versions of the model (the one that's in prod right now for example), so this `if` is needed for backward compatibility.

lookout/style/typos/generation.py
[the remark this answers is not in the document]
zurk: I mean if they put `load_facebook_vectors` if there any other? 🤔 nevermind.

lookout/style/typos/generation.py
zurk: but why do you delete `self.min_freq`? in other conditions, it can be vice-versa. Let's just set to 0.

lookout/style/typos/generation.py
zurk: do we repeat it twice before?

lookout/style/typos/generation.py
zurk: I see that you set `buckets_word` to None if there is no `buckets_word` does it mean that it is not necessary and model can work without `buckets_word`?

lookout/style/typos/generation.py
irinakhismatullina: Yep, it's calculated automatically (and probably very cheap too) during the `checker.lookup` for the returned tokens. And then it was calculated separately during the features calculations, so not cool.

lookout/style/typos/generation.py
irinakhismatullina: Ok, will do that

lookout/style/typos/generation.py
irinakhismatullina: Yes, actually it was empty in the last tried versions of the fasttext models. And when I wrote this code a couple of months ago and ran with the previous version of gensim, it was not empty. 
It may be working like some kind of a cache to store stuff now, but I'm not sure.

Tbh it looks like they (gensim) are currently slightly lost in fasttext model storage logic, they have complex inheritance architecture, so it's not ideal. I'm trying to keep all the attributes stored just in case.

zurk: let's find out if we need `buckets_word ` or remove it.

lookout/style/typos/generation.py
zurk: all right.

lookout/style/typos/generation.py
irinakhismatullina: I don't know the architecrute of the gensim fasttext so well and I don't think it's reasonable to waste time on learning it. Imo we should always save all the attributes.

zurk: ok, then let's remove it and upload the new fixed model.

irinakhismatullina: I don't get what you mean

zurk: remove what is related to old model backward compatibility and upload a new one, please.

irinakhismatullina: This way I will just remove the `if "buckets_word" in self.wv.keys()` part, and anyway I think we should have backward compatibility. At least while I can't acess the new model.

zurk: ok, please add a comment then.

lookout/style/typos/generation.py
```diff
@@ -280,7 +288,7 @@ def _generate_features(self, typo_info: TypoInfo, dist: int, typo_vec: numpy.nda
                 self._min_cos(candidate_vec, context),
                 self._cos(typo_vec, candidate_vec),
                 dist,
-                int(candidate in self.tokens),
+                1.0 if dist > 0 else float(candidate in self.tokens),
```
vmarkovtsev:
```suggestion
                float(dist > 0 or candidate in self.tokens),
```

lookout/style/typos/generation.py
```diff
@@ -375,18 +383,26 @@ class DummyModel(Model):
         for key, val in self.wv.vocab.items():
             vocab_strings[val.index] = key
             vocab_counts[val.index] = val.count
-        hash2index = numpy.zeros(len(self.wv.hash2index), dtype=numpy.uint32)
-        for key, val in self.wv.hash2index.items():
-            hash2index[val] = key
+        if isinstance(self.wv.buckets_word, dict):
```
vmarkovtsev: How come we have an "if" here?

lookout/style/typos/generation.py
```diff
@@ -419,9 +435,16 @@ def _load_tree(self, tree: dict) -> None:
             for i, s in enumerate(vocab)}
         wv.bucket = self.wv["bucket"]
         wv.index2word = wv.index2entity = vocab
-        wv.num_ngram_vectors = self.wv["num_ngram_vectors"]
         wv.vectors_ngrams = numpy.array(self.wv["vectors_ngrams"])
-        wv.hash2index = {k: v for v, k in enumerate(self.wv["hash2index"])}
+        # This if check is needed for supporting models that were saved with gensim < 3.7.2
+        if "buckets_word" in self.wv.keys():
```
vmarkovtsev:
```suggestion
        if "buckets_word" in self.wv:
```

lookout/style/typos/generation.py
irinakhismatullina: It sometimes can be a dict, and sometimes a NoneType.

When I wrote this code a couple of months ago with `gensim==3.7.2`, there was a dict in the model, loaded with `load_facebook_vectors`. RIght now there's `NoneType`, but probably it can become `dict` when smth will be cached, I'm not sure.

irinakhismatullina: Ok, this attribute seem to not be used when the model is loaded from the facebook dump, so I removed it.

zurk:
CI :)

zurk:
Vadim's feedback was addressed => merging.

